PR #3, opened 2016-09-14
Readme and git organizational changes

jrheizelman:
Added a line in the main repo readme to list the framework.

Moved around .gitignore and .gitmodules files to live in the main directory rather than where they formerly were in each project's folder.


jrheizelman:
It looks like this pull request is including all of the changes of the last pull request as well, even though they have been merged into the main repo already. Not sure what the problem is, reaching out to Nico to see if he can see what mistake we made.

jrheizelman:
After advice from Nico, I rebased the branch on my fork, which solved the issue. Now only the minimal changes I made most recently are included in this pull request.

README.md
```diff
@@ -4,3 +4,6 @@ available are:
 
 * [Kafka Connector](https://github.com/GoogleCloudPlatform/pubsub/tree/master/kafka-connector):
   Send and receive messages from [Apache Kafka](http://kafka.apache.org).
+* [Load Testing Framework](https://github.com/GoogleCloudPlatform/pubsub/tree/master/load-test-framework):
+  Set up comparative load tests between [Apache Kafka](http://kafka.apache.org)
+  and [Google Cloud Pub/Sub](https://cloud.google.com/pubsub/).
```
aspicool: Also include two clients of the same stack. For example Http/Json and gRPC clients of CPS.

jrheizelman:
Travis tests not set up yet, so the build failure doesn't reflect a failed test. Was starting to integrate Travis in another branch, so this check doesn't mean anything.

I also changed the Travis settings so these attempts to run tests will no longer occur when there aren't tests set up, which should stop these warning signs until we get Travis up and running.

PR #4, opened 2016-09-19
Added Travis testing w/ build changes, and fixed CPS subscription

jrheizelman:
This doesn't have any actual tests included yet, just makes sure the build works for now. In the future, I'll add some as I start writing.


jrheizelman:
Finished working on CPS subscription to get it working even when trying to reuse a subscription name that already exists for the topic. In this case, originally I tried to get the old subscription and pull messages down, but that led to incorrect results due to old messages that hadn't been pulled down yet to come first, making the latency numbers potentially days. To fix, if the framework tries to create a subscription that already exists, it will delete the existing one and create a new one to be sure to only get messages sent after the load test started. 

I'll also add a something describing this to the readme, but I didn't think deleting the straggling old messages before running a load test with thousands of new meaningless ones was a big problem.

load-test-framework/src/main/java/com/google/pubsub/flic/cps/CPSRoundRobinSubscriber.java
```diff
@@ -61,4 +62,11 @@ public CPSRoundRobinSubscriber(int numClients) throws Exception {
   public ListenableFuture<Subscription> createSubscription(Subscription request) {
     return pullClients.get(0).createSubscription(request);
   }
+  
+  /** Delete a Subscription. */
+  public ListenableFuture<Empty> deleteSubscription(Subscription request) {
+    DeleteSubscriptionRequest.Builder deleteBuilder = DeleteSubscriptionRequest.newBuilder();
+    deleteBuilder.setSubscription(request.getName());
+    return pullClients.get(0).deleteSubscription(deleteBuilder.build());
```
mdietz94: Nit: I think the following is more idiomatic
```
return pullClients.get(0).deleteSubscription(
  DeleteSubscriptionRequest.newBuilder()
    .setSubscription(request.getName())
    .build());
```

load-test-framework/src/main/java/com/google/pubsub/flic/cps/CPSSubscribingTask.java
```diff
@@ -70,9 +74,25 @@ public void execute() throws Exception {
     AcknowledgeRequest.Builder ackBuilder = AcknowledgeRequest.newBuilder();
     // Create a subscription for each topic.
     for (String topic : args.getTopics()) {
-      Subscription request =
-          subBuilder.setTopic(Utils.getCPSTopic(topic, args.getCPSProject())).build();
-      Subscription response = subscriber.createSubscription(request).get();
+      String[] splitOnColon = topic.split(":");
+      if(splitOnColon.length != 2)
+        throw new ParameterException("Parameter topics should be formatted 'topic:subscription'");
+      subBuilder.setName(Utils.getCPSSubscription(splitOnColon[1], args.getCPSProject()));
```
mdietz94: This should be (builder functions return a builder):
```
subBuilder.setName(...)
          .setTopic(...);
```

load-test-framework/src/main/java/com/google/pubsub/flic/cps/CPSRoundRobinSubscriber.java
jrheizelman: Changed this in more recent commit

load-test-framework/src/main/java/com/google/pubsub/flic/cps/CPSSubscribingTask.java
jrheizelman: Changed this in more recent commit

jrheizelman:
I made the builder changes you suggested in the recent commit to match the convention better.

PR #5, opened 2016-09-22
Added stats tracking to Kafka publishing tasks

jrheizelman:


load-test-framework/src/main/java/com/google/pubsub/flic/processing/MessageProcessingHandler.java
```diff
@@ -140,14 +140,16 @@ public void printStats(long start, DelayTrackingThreadPool executor, AtomicBoole
...
                 latencyStats.getValueAtPercentile(99)));
-    log.info(
-        "The average delay for processing "
-            + latencyType
-            + " latency was "
-            + executor.getAverageTaskWaitTime()
-            + ""
-            + " "
-            + "ms");
+    if(executor != null) {
```
mdietz94: Should executor ever be null? You should use an annotation above for @Nullable if so.

jrheizelman: Passing in null was a temporary fix because the Kafka publishing task code wasn't tracking the tasks in the same way, more relying on the Apache client to do the work for it. Because it wasn't updating the executor or using it at all, passing in null was easier. After thinking about it more, I realized that the way it's set up now doesn't give a 1:1 comparison. On the CPS side, the publisher tracks the time up until when the batch is sent off, but the Kafka publishing code doesn't do this, it just sets the start when it first starts sending any messages, so that batch->send delay is built-in against Kafka. With a large load test, that delay could be pretty significant, so I think it's best for there to be no data rather than bad data and hold off on putting this in.

I'll work on those changes and let you know when this PR is ready again.

jrheizelman:
This turned out to be easier than I thought after looking into the Kafka Producer code. The Apache client does a lot of the batching stuff on its own and sets timestamps accordingly. I also edited the readme, because a lot of the features we talked about not existing in our implementation are actually present in the library class we're using. 

Eventually it might be good to do something like create a child of the Apache class to better set up internal data tracking and integrating its controls with command line args, but we can cross that bridge if we need to.

PR #6, opened 2016-09-28
Ability to Orchestrate Clients on GCE

mdietz94:
This pull requests adds functionality to initialize a project for use by the loadtest framework, and for bringing up and shutting down instances on Google Compute Engine. This is part of a larger change to make this framework more flexible.

Also contains some small optimizations and lint fixes to existing code.


mdietz94:
AFAICT, the build failure is transient and unrelated to me. It failed downloading a Kafka jar, and my brach built successfully.

jrheizelman:
Haven't looked at code at all, but the build has passed now, so that looks fine.

load-test-framework/src/main/resources/command.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
```
jrheizelman: It came up somewhere in this directory, I think, where the Command.java file (the really long, do not edit file produced by the protobuf compiler). That java file is automatically produced with the maven build right? Could we remove it from the directory and add to .gitignore just to clean up the repo a little bit?

load-test-framework/src/main/resources/command.proto
```diff
@@ -0,0 +1,23 @@
...
+
+option java_package = "com.google.pubsub.flic.common";
+option java_outer_classname = "Command";
+
+service LoadtestFramework {
+  rpc StartClient(CommandRequest) returns (google.protobuf.Empty) {
+  }
+}
+
+message CommandRequest {
```
mdietz94: The tasks will be different clients. So the instance started will already be a publisher / subscriber, and these are the arguments that will be given. (I.e. there will be startup scripts that provide initial arguments, like subscribe, publish, etc. that are defining of the type of client. We could put that in the command instead of the script, there's no material difference, but it seems easier since then each instance group / template can be separated by publish/subscribe and type. We should loop up tomorrow to discuss further. This is definitely not the full list of options, but it was my first pass.

load-test-framework/src/main/resources/command.proto
mdietz94: Correct, good catch, thanks.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,331 @@
...
+  private static final Logger log = LoggerFactory.getLogger(GCEController.class.getName());
+  private final Storage storage;
+  private final Compute compute;
+  private final Executor executor;
+  private List<ClientType> types;
+  private boolean shutdown;
+  private final String machineType = "n1-standard-4"; // quad core machines
+  private final String sourceFamily = "projects/debian-cloud/global/images/family/debian-8"; // latest Debian 8
+  private final String projectName;
+  private final String zone = "us-central1-a";
```
mdietz94: This is just a placeholder since as of right now this is just a single zone. Eventually I'm going to make the constructor argument
Map<Zone, Map<ClientType, Integer>> and then we can have arbitrary numbers of client types in different zones that all work. My plan is to do that in another PR (this one seemed long enough to me already, but I can do it in this one if you feel strongly about it).

load-test-framework/src/main/resources/command.proto
jrheizelman: Yeah I like that better after thinking about it. More like real-usage for the publisher and subscriber to be different instances.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
jrheizelman: That sounds good, and it can definitely wait. Mostly just wanted to make sure it was on your radar.

load-test-framework/src/main/proto/command.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+import "google/protobuf/empty.proto";
+import "google/protobuf/timestamp.proto";
+package pubsub;
+
+option java_package = "com.google.pubsub.flic.common";
+option java_outer_classname = "Command";
+
+service LoadtestFramework {
+  rpc StartClient(CommandRequest) returns (google.protobuf.Empty) {
```
amordkov: Comment/doc?

amordkov: Does "command" have any special meaning? Why is the file called command.proto? Generally, the name of the request message should be <method_name>Request. So here it would StartClientRequest. And the response should be StartClientResponse -- it can be empty for now, but we may want to add fields to it later.

load-test-framework/src/main/proto/command.proto
```diff
@@ -0,0 +1,23 @@
...
+  }
+}
+
+message CommandRequest {
+  google.protobuf.Timestamp start_time = 1;
```
amordkov: Please add comments/documentation to these fields. In particular, please address these points:
- What happens if start_time is in the past? What about stop_time?
- What happens if start_time >= stop_time?

mdietz94: If start_time is in the past, we start immediately. If stop_time is in the past, we stop immediately. So if they are both in the past we will basically just immediately start then stop.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+      description = "The network address of the Kafka broker."
+  )
+  private String broker;
+  @Parameter(
+      names = {"--request_rate"},
+      description = "The rate at which each client will make requests."
+  )
+  private int requestRate = 10;
+  @Parameter(
+      names = {"--max_concurrent_requests"},
```
amordkov: s/concurrent/outstanding/g

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+  )
+  private String project = "";
+  @Parameter(
+      names = {"--batch_size", "-b"},
+      description = "Number of messages to batch per publish request.",
+      validateWith = GreaterThanZeroValidator.class
+  )
+  private int batchSize = 10;
+  @Parameter(
+      names = {"--subscriber_fanout"},
```
amordkov: This is the number of subscriptions attached to each topic, right? If so, suggest "subscription_fanout".

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+      description = "Number of subscription ids to use for each topic. Must be at least 1.",
```
amordkov: "Number of subscriptions to create for each topic"

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+  )
+  private boolean help = false;
+  @Parameter(
+      names = {"--cps_publisher_count"},
+      description = "Number of CPS publishers to start."
+  )
+  private int cpsPublisherCount = 1;
+  @Parameter(
+      names = {"--cps_subscriber_count"},
+      description = "Number of CPS subscribers to start per ."
```
amordkov: per what?

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+  @Parameter(
+      names = {"--loadtest_seconds"},
+      description = "Duration of the load test, in seconds.",
+      validateWith = GreaterThanZeroValidator.class
+  )
+  private int loadtestLengthSeconds = 60;
+  @Parameter(
+      names = {"--project"},
+      required = true,
+      description = "Cloud Pub/Sub project name."
```
amordkov: "GCP project name"

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+  )
+  private int maxConcurrentRequests = 20;
+  @Parameter(
+      names = {"--burn_in_duration_seconds"},
+      description = "The duration, in seconds, to run without recording statistics in order to allow tuning."
+  )
+  private int burnInDurationSeconds = 20;
+  @Parameter(
+      names = {"--number_of_messages"},
+      description = "The number of messages to publish in the test. Enabling this will override " +
```
amordkov: "total number of messages"?

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
-          jCommander.usage();
-        }
-        return;
-      }
-      if (jCommander.getParsedCommand().equals(CompareArguments.COMMAND)) {
-        // Compares data dumps for correctness.
-        Comparison c = new Comparison(dataComparisonArgs.getFile1(), dataComparisonArgs.getFile2());
-        c.compare();
-        return;
+      Map<String, Map<ClientParams, Integer>> clientTypes = ImmutableMap.of(
```
amordkov: Just build up a Map\<ClientParams, Integer\> clientParamsMap first and then pass a:

ImmutableMap.of( "us-central1-a", clientParamsMap)

at line 177 below.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+          cpsPublisherCount > 0 ||
+              cpsSubscriberCount > 0 ||
+              kafkaPublisherCount > 0 ||
+              kafkaSubscriberCount > 0
+      );
+      Preconditions.checkArgument(
+          broker != null || (kafkaPublisherCount == 0 && kafkaSubscriberCount == 0));
+      for (int i = 0; i < subscriberFanout; ++i) {
+        clientTypes.get("us-central1-a").put(new ClientParams(ClientType.CPS_GCLOUD_PUBLISHER, null),
+            cpsPublisherCount / subscriberFanout);
```
amordkov: This is awkward (and not precise if cpsPublisherCount is not divisible by subscriberFanout). Can we just just start cpsPublisherCount publishers in a separate loop?

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+        clientTypes.get("us-central1-a").put(new ClientParams(ClientType.CPS_GCLOUD_SUBSCRIBER,
+                "gcloud-subscription" + i),
+            cpsSubscriberCount / subscriberFanout);
```
amordkov: Is the assumption that there is only a single topic? If so, then yeah there's going to be subscriberFanout subscriptions and this divides the cpsSubscriberCount subscribers among all the subscriptions. But that is not clear.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+        clientTypes.get("us-central1-a").put(new ClientParams(ClientType.KAFKA_PUBLISHER, null),
+            kafkaPublisherCount / subscriberFanout);
```
amordkov: Ditto; this should be done in a separate kafkaPublisherCount-based loop.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+        clientTypes.get("us-central1-a").put(new ClientParams(ClientType.KAFKA_SUBSCRIBER, null),
+            kafkaSubscriberCount / subscriberFanout);
```
amordkov: Ditto; this should be done in a separate kafkaSubscriberCount-based loop.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,122 +15,224 @@
...
+      Client.loadtestLengthSeconds = loadtestLengthSeconds;
+      Client.batchSize = batchSize;
+      Client.broker = broker;
+      Client.requestRate = requestRate;
+      Client.maxConcurrentRequests = maxConcurrentRequests;
+      Client.burnInTimeMillis = (Client.startTime.getSeconds() + burnInDurationSeconds) * 1000;
+      Client.numberOfMessages = numberOfMessages;
+      GCEController gceController =
+          GCEController.newGCEController(project, clientTypes, Executors.newScheduledThreadPool(500));
+      gceController.initialize();
```
amordkov: Can the initialize() work just be done in the constructor?

load-test-framework/src/main/java/com/google/pubsub/flic/common/LatencyDistribution.java
```diff
@@ -0,0 +1,117 @@
...
+
+    boolean bucketFound = false;
+    for (int i = 0; i < LATENCY_BUCKETS.length; i++) {
+      double bucket = LATENCY_BUCKETS[i];
+      if (latencyMs < bucket) {
+        synchronized (this) {
+          bucketValues[i]++;
+        }
+        bucketFound = true;
+        break;
```
amordkov: Just return here. That way you don't need the bucketFound variable.

load-test-framework/src/main/java/com/google/pubsub/flic/common/LatencyDistribution.java
```diff
@@ -0,0 +1,117 @@
...
+      }
+    }
+    if (!bucketFound) {
```
amordkov: This will happen only if latencyMs is Integer.MAX_VALUE, right?

load-test-framework/src/main/proto/command.proto
mdietz94: Fixed

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -0,0 +1,73 @@
...
+
+  // The length of time to poll for.
+  int32 poll_length = 2;
+}
+
+message CheckRequest {
+}
+
+message CheckResponse {
+  repeated int64 bucket_values = 1;
```
mdietz94: These are the bucket values from the LatencyDistribution. This is the same name that we use in that class, and the same name that Stackdriver uses in their API for DistributionValues.

load-test-framework/src/main/java/com/google/pubsub/clients/common/LoadTestRunner.java
```diff
@@ -0,0 +1,134 @@
...
+  private static Task client;
+  private static AtomicBoolean finished = new AtomicBoolean(false);
+  private static SettableFuture<Void> finishedFuture = SettableFuture.create();
+
+  public static void run(Function<StartRequest, Task> loadFunction) throws Exception {
+    SettableFuture<StartRequest> requestFuture = SettableFuture.create();
+    server = ServerBuilder.forPort(5000)
+        .addService(new LoadtestGrpc.LoadtestImplBase() {
+          @Override
+          public void start(StartRequest request, StreamObserver<StartResponse> responseObserver) {
```
mdietz94: It is not. StreamObserver is what is used for all gRPC calls in Java. The difference is that for non-streaming RPCs it is an error to call onNext more than once.

load-test-framework/src/main/java/com/google/pubsub/clients/common/MetricsHandler.java
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  private void reportMetrics() {
+    CreateTimeSeriesRequest request;
+    synchronized (this) {
+      String now = dateFormatter.format(new Date());
+      request = new CreateTimeSeriesRequest().setTimeSeries(Collections.singletonList(
+          new TimeSeries()
+              .setMetric(new Metric()
+                  .setType("custom.googleapis.com/cloud-pubsub/loadclient/" + metricName)
```
mdietz94: This can be anything that starts with custom.googleapis.com and that does not already exist in your project.

load-test-framework/src/main/java/com/google/pubsub/clients/common/MetricsHandler.java
```diff
@@ -0,0 +1,222 @@
...
+      distribution.reset();
+    }
+    try {
+      monitoring.projects().timeSeries().create("projects/" + project, request).execute();
+    } catch (IOException e) {
+      log.error("Error reporting latency.", e);
+    }
+  }
+
+  List<Long> getBucketValues() {
```
mdietz94: No, you need to know the exact buckets, but everything is linked against LatencyDistribution, so it would be impossible to do anything wrong here. The alternative is passing around an entire DistributionValue object, but that would require checking the buckets match up, which seems excessively complicated since we control all the code and only define a single set of buckets.

load-test-framework/src/main/java/com/google/pubsub/clients/gcloud/CPSSubscriberTask.java
```diff
@@ -0,0 +1,76 @@
...
+        .build().service();
+    this.subscription = Preconditions.checkNotNull(subscription);
+    this.batchSize = batchSize;
+  }
+
+  public static void main(String[] args) throws Exception {
+    LoadTestRunner.run(request ->
+        new CPSSubscriberTask(request.getProject(), request.getPubsubOptions().getSubscription(),
+            request.getPubsubOptions().getMaxMessagesPerPull(),
+            request.getStopConditionsCase() == StopConditionsCase.NUMBER_OF_MESSAGES ?
```
mdietz94: Correct, I forgot to revert this while I was working. Fixed.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Fixed. Was going to say zone, but that feature is not yet done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Fixed.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Fixed.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Yeah did that, not sure what I was thinking here.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Correct. That is built in, as users cannot set topics. This seems like something I should add to the Readme?

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Removed, fanout doesn't seem to exist in Kafka, seems like you would need to publish to multiple topics for replication.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Yes. Fixed.

load-test-framework/src/main/java/com/google/pubsub/flic/common/LatencyDistribution.java
mdietz94: Fixed.

load-test-framework/src/main/java/com/google/pubsub/flic/common/LatencyDistribution.java
mdietz94: Correct. By Stackdriver rules you need an underflow and overflow bucket.

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -0,0 +1,73 @@
...
+  int32 request_rate = 3;
+
+  // The size of each user message to publish
+  int32 message_size = 4;
+
+  // The number of user messages of size message_size to batch together in each publish
+  int32 batch_size = 5;
+
+  // The maximum concurrent requests, per client.
+  int32 max_concurrent_requests = 6;
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Fixed.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Fixed.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Fixed.

load-test-framework/src/main/java/com/google/pubsub/clients/common/LoadTestRunner.java
```diff
@@ -0,0 +1,134 @@
...
+            if (requestFuture.isDone()) {
+              responseObserver.onError(new Exception("Start should only be called once, ignoring this request."));
+              return;
+            }
+            requestFuture.set(request);
+            responseObserver.onNext(StartResponse.getDefaultInstance());
```
mdietz94: Then we'll fail the health check. Due to the start time being a delay, we can't have an RPC open long enough to actually make sure we can start successfully.

amordkov: Alright, so I guess there are no "synchronous" errors that we can return? Like, if input validation fails?

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,222 @@
...
-          jCommander.usage();
-        }
-        return;
-      }
-      if (jCommander.getParsedCommand().equals(CompareArguments.COMMAND)) {
-        // Compares data dumps for correctness.
-        Comparison c = new Comparison(dataComparisonArgs.getFile1(), dataComparisonArgs.getFile2());
-        c.compare();
-        return;
+      //Map<String, Map<ClientParams, Integer>> clientTypes = ImmutableMap.of(
```
amordkov: dead code?

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
amordkov: Yes, and also add a comment here.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -0,0 +1,239 @@
...
+      requestBuilder.setNumberOfMessages(numberOfMessages);
+    } else {
+      requestBuilder.setStopTime(Timestamp.newBuilder()
+          .setSeconds(Math.max(startTime.getSeconds(), System.currentTimeMillis() / 1000) +
+              loadtestLengthSeconds).build());
+    }
+    switch (clientType) {
+      case CPS_GCLOUD_SUBSCRIBER:
+        requestBuilder.setPubsubOptions(PubsubOptions.newBuilder()
+            .setMaxMessagesPerPull(10)
```
amordkov: Shouldn't this be configured by inputs?

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -0,0 +1,239 @@
...
+            .setSubscription(subscription));
+        break;
+      case KAFKA_PUBLISHER:
+      case KAFKA_SUBSCRIBER:
+        requestBuilder.setKafkaOptions(KafkaOptions.newBuilder()
+            .setBroker(broker)
+            .setPollLength(100));
```
amordkov: Shouldn't this be configured by inputs?

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -0,0 +1,239 @@
...
+  public static int messageSize;
+  public static int requestRate;
+  public static Timestamp startTime;
+  public static int loadtestLengthSeconds;
+  public static int batchSize;
+  public static String broker;
+  public static int maxOutstandingRequests;
+  public static long burnInTimeMillis;
+  public static int numberOfMessages = 0;
+  private final ClientType clientType;
```
amordkov: I'm confused as to why some of these settings are static and some are not. How did you decide?

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -0,0 +1,239 @@
...
+        } catch (InterruptedException e) {
+          log.info("Interrupted during back off, retrying.");
+        }
+        log.debug("Going to retry client connection, likely due to start up time.");
+        stub = getStub();
+        stub.start(request, this);
+      }
+
+      @Override
+      public void onCompleted() {
```
amordkov: What's the difference between onNext() and onCompleted() for non-streaming? It seems like both startFuture.set(null) and clientStatus = ClientStatus.RUNNING should be done in the same callback.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -0,0 +1,239 @@
...
+        if (checkResponse.getIsFinished()) {
+          clientStatus = ClientStatus.STOPPED;
+          doneFuture.set(null);
+        }
+        if (System.currentTimeMillis() < burnInTimeMillis) {
+          return;
+        }
+        synchronized (this) {
+          for (int i = 0; i < LatencyDistribution.LATENCY_BUCKETS.length; i++) {
+            bucketValues[i] += checkResponse.getBucketValues(i);
```
amordkov: I'm not sure that it's correct to add on the bucket values. The server returns cumulative state, right? Seems like that would result in a lot of double-counting of events.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -0,0 +1,239 @@
...
+          }
+        }
+      }
+
+      @Override
+      public void onError(Throwable throwable) {
+        if (errors > 3) {
+          clientStatus = ClientStatus.FAILED;
```
amordkov: Which client is this supposed to be the status of? The thing running locally or the load-generating thing in GCE? I think it's the latter, but that's confusing because this file is called Client.java. Also, if it's the latter, I don't think we're differentiating between:

- Failure to reach the server
- The load generation completing successfully
- The load generation completing prematurely with an error.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.stream.Collectors;
+
+public abstract class Controller {
+  final static Logger log = LoggerFactory.getLogger(Controller.class);
+  final List<Client> clients = new ArrayList<>();
+  final ScheduledExecutorService executor;
+
+  /*
+  Creates the given environments and starts the virtual machines. When this function returns, each client is guaranteed
```
amordkov: plz make proper javadoc

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
```diff
@@ -0,0 +1,122 @@
...
+  to have been connected and be network reachable, but is not started. If an error occurred attempting to start the
+  environment, the environment will be shut down, and an IOException will be thrown. It is not guaranteed that we have
```
amordkov: Is this still true?

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
```diff
@@ -0,0 +1,122 @@
...
+            .collect(Collectors.toList())
+        ).get();
+      } catch (ExecutionException e) {
+        throw e.getCause();
+      }
+      result.endTimeMillis = System.currentTimeMillis();
+    }
+    clientsOfType.stream().map(Client::getBucketValues).forEach(bucketValues -> {
+      for (int i = 0; i < LatencyDistribution.LATENCY_BUCKETS.length; i++) {
+        result.bucketValues[i] += bucketValues[i];
```
amordkov: Ditto re: accumulating bucket values.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
```diff
@@ -0,0 +1,122 @@
...
+  environment, the environment will be shut down, and an IOException will be thrown. It is not guaranteed that we have
+  completed shutting down when this function returns, but it is guaranteed that we are in process.
+   */
+  Controller(ScheduledExecutorService executor) {
+    this.executor = executor;
+  }
+
+  protected abstract void shutdown(Throwable t);
+
+  private Result resultsForType(Client.ClientType type, boolean wait) throws Throwable {
```
amordkov: Instead of having a wait parameter and calling now() on the local controller (which may be different from the clock on the remote VM), the VM server should respond with a running_time in the CheckResponse (which is independent of the 'done' bit in the CheckResponse). This way, the throughput calculation will be very simple and getResults can be invoked in the same way, regardless of whether the VM is done generating its load.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Yes, fixed.

load-test-framework/src/main/proto/command.proto
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,222 @@
...
+  private String project = "";
+  @Parameter(
+      names = {"--batch_size", "-b"},
+      description = "Number of messages to batch per publish request.",
+      validateWith = GreaterThanZeroValidator.class
+  )
+  private int batchSize = 10;
+  @Parameter(
+      names = {"--subscription_fanout"},
+      description = "Number of subscriptions to create for each topic. Must be at least 1.",
```
mdietz94: Done. Multiple topics will be used when we have multiple clients. When we want to load test the Python library vs the Java library each will use a separate topic. But this is CPS only.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Yes, Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,222 @@
...
+          broker != null || (kafkaPublisherCount == 0 && kafkaSubscriberCount == 0));
+      Map<ClientParams, Integer> clientParamsMap = new HashMap<>();
+      clientParamsMap.putAll(ImmutableMap.of(
+          new ClientParams(ClientType.CPS_GCLOUD_PUBLISHER, null), cpsPublisherCount,
+          new ClientParams(ClientType.KAFKA_PUBLISHER, null), kafkaPublisherCount,
+          new ClientParams(ClientType.KAFKA_SUBSCRIBER, null), kafkaSubscriberCount
+      ));
+      for (int i = 0; i < subscriptionFanout; ++i) {
+        clientParamsMap.put(new ClientParams(ClientType.CPS_GCLOUD_SUBSCRIBER, "gcloud-subscription" + i),
+            cpsSubscriberCount / subscriptionFanout);
```
mdietz94: Done, documented. Seems worse for some subscriptions to have more subscribers than others.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,222 @@
...
-          }
-          taskArgs = builder.broker(kafkaArgs.getBroker()).build();
-          kafkaHandler.setLatencyType(MessageProcessingHandler.LatencyType.END_TO_END);
-          KafkaConsumer<String, String> consumer =
-              KafkaConsumerTask.getInitializedConsumer(taskArgs);
-          log.info("Creating a task which consumes from Kafka.");
-          new KafkaConsumerTask(taskArgs, consumer, kafkaHandler).execute();
+      Client.messageSize = messageSize;
+      Client.requestRate = 1;
+      Client.startTime = Timestamp.newBuilder().setSeconds(System.currentTimeMillis() / 1000 + 90).build();
```
mdietz94: Currently it is just a safe default, since 90 seconds is enough time to guarantee everything has started with a 30 second or so buffer.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
mdietz94: Static settings are those that will always be the same for all clients. Others are not static. Project may eventually be different when we allow orchestration on other environments.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
mdietz94: It does not return cumulative state, but deltas. This is because Stackdriver does not plot cumulative values. Also it will later allow us to calculate current throughput.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
mdietz94: The latter. We are. If it completes successfully it lets us know and we do not keep checking it. We do not differentiate between 1 and 3 but both are considered an error. Also 1 should only happen as a result of 3. We retry three times in order to make a transient failure to reach the server not affect the load test. If you have a better name in mind for this class, let me know.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
mdietz94: Yes. That is why I moved it to the constructor.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
mdietz94: Same answer as above, this is correct. Would have done cumulative if Stackdriver would accept it.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
amordkov: Alternatively, instead of a cpsSubscribersCount flag, you could have a cpsSubscribersPerSubscription flag.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,236 @@
...
+    results.forEach((type, result) -> {
+      log.info("Results for " + type + ":");
+      log.info("50%: " + LatencyDistribution.getNthPercentile(result.bucketValues, 0.5));
+      log.info("99%: " + LatencyDistribution.getNthPercentile(result.bucketValues, 0.99));
+      log.info("99.9%: " + LatencyDistribution.getNthPercentile(result.bucketValues, 0.999));
+      // CPS Publishers report latency per batch message.
+      log.info("Average throughput: " +
+          new DecimalFormat("#.##").format(
+              (double) LongStream.of(
+                  result.bucketValues).sum() / (result.endTimeMillis / 1000.0 - Client.startTime.getSeconds())
```
amordkov: Do we still need to rely on Client.startTime if we have running_time from the CheckResponse?

mdietz94: Good catch. Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,235 @@
...
+    } catch (Throwable t) {
+      log.error("An error occurred...", t);
       System.exit(1);
     }
   }
+
+  private void printStats(Map<ClientType, Controller.Result> results) {
+    results.forEach((type, result) -> {
+      log.info("Results for " + type + ":");
+      log.info("50%: " + LatencyDistribution.getNthPercentile(result.bucketValues, 0.5));
```
amordkov: nit: percentile parameters should be between 0 and 100, not 0 and 1.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,235 @@
...
+  )
+  private int loadtestLengthSeconds = 60;
+  @Parameter(
+      names = {"--project"},
+      required = true,
+      description = "Google Cloud Platform project name."
+  )
+  private String project = "";
+  @Parameter(
+      names = {"--batch_size", "-b"},
```
amordkov: --publish_batch_size

(but see comment below)

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,235 @@
...
+      log.info("Results for " + type + ":");
+      log.info("50%: " + LatencyDistribution.getNthPercentile(result.bucketValues, 0.5));
+      log.info("99%: " + LatencyDistribution.getNthPercentile(result.bucketValues, 0.99));
+      log.info("99.9%: " + LatencyDistribution.getNthPercentile(result.bucketValues, 0.999));
+      // CPS Publishers report latency per batch message.
+      log.info("Average throughput: " +
+          new DecimalFormat("#.##").format(
+              (double) LongStream.of(
+                  result.bucketValues).sum() / result.runningSeconds * messageSize / 1000000.0
+                  * (type == ClientType.CPS_GCLOUD_PUBLISHER ? batchSize : 1)) + " MB/s");
```
amordkov: If --batch_size applies only to cps then it should be --cps_publish_batch_size

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
```diff
@@ -0,0 +1,125 @@
...
+          resultFuture.set(null);
+        } catch (Throwable t) {
+          resultFuture.setException(t);
+        }
+      });
+    }
+    try {
+      Futures.allAsList(resultFutures).get();
+    } catch (ExecutionException | InterruptedException e) {
+      log.error("Client failed health check, will print results accumulated during test up to this point.",
```
amordkov: The "will print results accumulated during test up to this point" is unnecessary (and could also be inaccurate if the caller changes behavior)

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+        .distinct().forEach((clientType) -> {
+          SettableFuture<Void> pubsubFuture = SettableFuture.create();
+          pubsubFutures.add(pubsubFuture);
+          executor.execute(() -> {
+            // Delete each topic and subscription if it exists and create it new to avoid potential backlog from previous runs
+            String topic = Client.topicPrefix + Client.getTopicSuffix(clientType);
+            try {
+              pubSub.create(TopicInfo.of(topic));
+            } catch (Exception e) {
+              log.info("Topic already exists, reusing.");
```
amordkov: That's not necessarily true here.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+            }
+            paramsMap.keySet().stream().filter((params) -> params.clientType == clientType && params.subscription != null)
+                .map((params) -> params.subscription).forEach((subscription) -> {
+              pubSub.deleteSubscription(subscription);
+              pubSub.create(SubscriptionInfo.of(topic, subscription));
```
amordkov: So we create the pubsub resources regardless of whether the load will be generated against pubsub or kafka, right? Do we need to create kafka resources up front as well then?

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+      shutdown(e.getCause());
+      throw e.getCause();
+    } catch (Exception e) {
+      shutdown(e);
+      throw e;
+    }
+  }
+
+  public static GCEController newGCEController(
+      String projectName, Map<String,
```
amordkov: weird whitespace

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -0,0 +1,75 @@
...
+  string topic = 2;
+
+  // The number of requests that can be made, each second, per client.
+  int32 request_rate = 3;
+
+  // The size of each user message to publish
+  int32 message_size = 4;
+
+  // The number of user messages of size message_size to batch together in each publish
+  int32 batch_size = 5;
```
amordkov: publish_batch_size

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -0,0 +1,75 @@
...
+
+  // The maximum outstanding requests, per client.
+  int32 max_outstanding_requests = 6;
+
+  // The time at which the load test should start. If this is less than the current time, we start immediately.
+  google.protobuf.Timestamp start_time = 7;
+
+  oneof stop_conditions {
+    // The time at which the load test should be stopped. If less than the current time, the load test will not run.
```
amordkov: "If less than current time or less than start_time, the load test will not run."
I guess in this case, status will immediately be set to 'done'?

load-test-framework/src/main/proto/loadtest.proto
amordkov: Add comment saying that this is a histogram of latencies.

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -0,0 +1,75 @@
...
+  int32 poll_length = 2;
+}
+
+message CheckRequest {
+}
+
+message CheckResponse {
+  repeated int64 bucket_values = 1;
+  google.protobuf.Duration running_duration = 2;
+  bool is_finished = 3;
```
amordkov: It's still not clear to me how to differentiate between a load test that ran successfully to completion and one that failed early. There are situations where the VM instance will still be up an running, responding to Check RPCs, but the load generation failed prematurely: e.g., if the broker could not be reached.

load-test-framework/src/main/proto/loadtest.proto
mdietz94: Done.

load-test-framework/src/main/proto/loadtest.proto
mdietz94: We could have the load generation just fail out if it has a series of consecutive failures?

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
mdietz94: Done.

load-test-framework/src/main/proto/loadtest.proto
mdietz94: Correct.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
mdietz94: Done, changed print to return.

load-test-framework/src/main/proto/loadtest.proto
mdietz94: But you see throughput when the test finishes, so if the broker cannot be reached you'll see that at the end. I do like the idea of the load framework not caring about the RPCs, because if we overload a Kafka broker that is relevant test information.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
mdietz94: Done.

mdietz94:
All authors are Googlers.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+                        ScheduledExecutorService executor, Storage storage, Compute compute, PubSub pubSub) throws Throwable {
+    super(executor);
+    this.shutdown = false;
+    this.projectName = projectName;
+    this.types = types;
+    this.storage = storage;
+    this.compute = compute;
+
+    List<SettableFuture<Void>> pubsubFutures = new ArrayList<>();
+    types.values().forEach((paramsMap) -> paramsMap.keySet().stream().map((params) -> params.clientType)
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+        .distinct().forEach((clientType) -> {
+          SettableFuture<Void> pubsubFuture = SettableFuture.create();
+          pubsubFutures.add(pubsubFuture);
+          executor.execute(() -> {
+            // Delete each topic and subscription if it exists and create it new to avoid potential backlog from previous runs
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+          .setName("cloud-loadtest-framework-firewall-rule")
+          .setDescription("A firewall rule to allow the driver to coordinate load test instances.")
+          .setAllowed(ImmutableList.of(
+              new Firewall.Allowed()
+                  .setIPProtocol("tcp")
+                  .setPorts(Collections.singletonList("5000"))));
+      try {
+        compute.firewalls().get(projectName, "cloud-loadtest-framework-firewall-rule").execute();
+        compute.firewalls().update(projectName, "cloud-loadtest-framework-firewall-rule", firewallRule).execute();
+      } catch (GoogleJsonResponseException e) {
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+        compute.firewalls().insert(projectName, firewallRule).execute();
+      }
+      List<SettableFuture<Void>> filesRemaining = new ArrayList<>();
+      Files.walk(Paths.get("src/main/resources/gce"))
+          .filter(Files::isRegularFile).forEach(filePath -> {
+        SettableFuture<Void> fileRemaining = SettableFuture.create();
+        filesRemaining.add(fileRemaining);
+        executor.execute(() -> {
+          try {
+            uploadFile(filePath);
```
mdietz94: As discussed, no it shouldn't.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+        });
+      });
+      List<SettableFuture<Void>> createGroupFutures = new ArrayList<>();
+      types.forEach((zone, paramsMap) -> paramsMap.forEach((param, n) -> {
+        SettableFuture<Void> createGroupFuture = SettableFuture.create();
+        createGroupFutures.add(createGroupFuture);
+        executor.execute(() -> {
+          try {
+            createManagedInstanceGroup(zone, param.clientType);
+            createGroupFuture.set(null);
```
mdietz94: see above

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+
+      // Everything is set up, let's start our instances
+      log.info("Starting instances.");
+      List<SettableFuture<Void>> resizingFutures = new ArrayList<>();
+      types.forEach((zone, paramsMap) -> paramsMap.forEach((type, n) -> {
+        SettableFuture<Void> resizingFuture = SettableFuture.create();
+        resizingFutures.add(resizingFuture);
+        executor.execute(() -> {
+          try {
+            startInstances(zone, type.clientType, n);
```
mdietz94: see above

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+        log.info("Bucket missing, creating a new bucket.");
+        try {
+          storage.buckets().insert(projectName, new Bucket()
+              .setName("cloud-pubsub-loadtest")).execute();
+        } catch (GoogleJsonResponseException e1) {
+          shutdown(e1);
+          throw e1;
+        }
+      }
+    } catch (IOException e) {
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+        try {
+          compute.instanceGroupManagers().insert(projectName, zone,
+              (new InstanceGroupManager()).setName("cloud-pubsub-loadtest-framework-" + type)
+                  .setInstanceTemplate("projects/" + projectName +
+                      "/global/instanceTemplates/cloud-pubsub-loadtest-instance-" + type)
+                  .setTargetSize(0))
+              .execute();
+          created = true;
+        } catch (GoogleJsonResponseException e1) {
+          if (!e1.getDetails().getErrors().get(0).getReason().equals("resourceNotReady")) {
```
mdietz94: Can't, it uses 400 which is also invalid argument.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,446 @@
...
+                .setAccessConfigs(Collections.singletonList(new AccessConfig()))))
+            .setMetadata(new Metadata()
+                .setItems(Collections.singletonList(new Metadata.Items()
+                    .setKey("startup-script-url")
+                    .setValue("https://storage.googleapis.com/cloud-pubsub-loadtest/" + type + "_startup_script.sh"))))
+            .setServiceAccounts(Collections.singletonList(new ServiceAccount().setScopes(
+                Collections.singletonList("https://www.googleapis.com/auth/cloud-platform")))));
+  }
+
+  @Override
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -0,0 +1,424 @@
...
+                        ScheduledExecutorService executor, Storage storage,
+                        Compute compute, PubSub pubSub) throws Throwable {
+    super(executor);
+    this.projectName = projectName;
+    this.types = types;
+    this.storage = storage;
+    this.compute = compute;
+
+    // For each unique type of CPS Publisher, create a Topic if it does not already exist, and then delete
+    // and recreate any subscriptions under it so that we do not have backlog from previous runs.
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
```diff
@@ -28,6 +28,11 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.stream.Collectors;
 
+/**
+ * Each subclass of Controller is responsible for instantiating and cleaning up a given environment. When an environment
```
mdietz94: Done. IntelliJ defaulted to 120 chars per line and I didn't notice :frowning:

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
```diff
@@ -74,6 +94,11 @@ private Result resultsForType(Client.ClientType type) {
     return result;
   }
 
+  /**
+   * Gets the results for all available types.
+   *
+   * @return the map from type to result, every type running is a valid key
+   */
   public Map<Client.ClientType, Result> getResults() {
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Controller.java
```diff
@@ -98,7 +123,11 @@ private Result resultsForType(Client.ClientType type) {
     return results;
   }
 
-  void startClients() {
+  /**
+   * Sends a LoadtestFramework.Start command to all clients to commence the load test. When this function returns it is
```
mdietz94: Done.

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -0,0 +1,80 @@
...
+
+  // The maximum outstanding requests, per client.
+  int32 max_outstanding_requests = 5;
+
+  // The time at which the load test should start. If this is less than the current time, we start immediately.
+  google.protobuf.Timestamp start_time = 6;
+
+  oneof stop_conditions {
+    // The time at which the load test should be stopped. If less than the current time, we will complete immediately.
+    google.protobuf.Timestamp stop_time = 7;
```
mdietz94: Agreed, done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -15,124 +15,238 @@
...
+      Preconditions.checkArgument(
+          broker != null || (kafkaPublisherCount == 0 && kafkaSubscriberCount == 0));
+      Map<ClientParams, Integer> clientParamsMap = new HashMap<>();
+      clientParamsMap.putAll(ImmutableMap.of(
+          new ClientParams(ClientType.CPS_GCLOUD_PUBLISHER, null), cpsPublisherCount,
+          new ClientParams(ClientType.KAFKA_PUBLISHER, null), kafkaPublisherCount,
+          new ClientParams(ClientType.KAFKA_SUBSCRIBER, null), kafkaSubscriberCount
+      ));
+      // Each type of client will have its own topic, so each subscription will get
+      // cpsSubscriberCount subscribers cumulatively among each of the subscriptions.
```
amordkov: I don't really understand this sentence. It's trying to say that there are cpsSubscriptionFanout subscriptions, and each subscription will have 'cpsSubscriberCount / cpsSubscriptionFanout' subscribers pulling from it, right?

PR #8, opened 2016-10-06
Stat aggregation from files, and misc changes to data tracking

jrheizelman:
The biggest change is a new command that pulls in the information dumped to files, aggregates the data, and prints the latency and throughput data. We'd included the data dump previously, but had no way of reading it once it was dumped. 

There are also other changes used for throughput calculation on subscriber tasks. Previously, the start time was when the subscription was opened, so if you started the subscriber task, then took some time before running the publisher, that time would be included in the denominator that calculated average throughput, making it artificially low. Now, to calculate total running time, it tracks the earliest received message to have a better start point.

There's also some cleanup in the gitignore, removing the proto-generated java file, and adding that compilation to maven pom.


mdietz94:
Since all the files are being removed and stat aggregation is being done live through RPCs now, throughput calculations are done correctly in the new version, so I don't think it's likely any of this will still be useful, but if so feel free to make these changes to the new version.

PR #11, opened 2016-10-14
Max Publish Latency and Subscribe Throughput Testing

mdietz94:
Allows tests to specify a max latency and increase the request rate until they hit that latency. Also allows servers to restart "hot" for faster testing.


load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -164,6 +193,11 @@ private void run() {
       );
```
amordkov: plz specify error message

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -141,6 +154,22 @@
...
+      description = "This sets the maximum latency, in this test we will continuously run load " +
+          "tests with increasing request rates until we hit the provided latency. You must only " +
+          "provide a single type of publisher to use this test. This uses the 99% latency as the " +
+          "bound to check."
+  )
+  private int maxPublishLatency = 0;
+  @Parameter(
+      names = {"--max_subscriber_throughput_test"},
+      description = "This test will continuously run load tests with greater publish request " +
+          "rate until the subscriber can no longer keep up, and will let you know the maximum " +
```
amordkov: Can we parameterize in a flag the "no longer keeping up" condition? Perhaps a size of backlog or an oldest unacked message age.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -141,6 +154,22 @@
           "message loss. If set less than 1, this flag is ignored."
   )
   private int numberOfMessages = 0;
+  @Parameter(
+      names = {"--max_publish_latency_test"},
```
amordkov: This flag name is confusing; it's unclear whether it's an int or a string. Suggest "max_publish_latency". However, I think the bigger problem is that this flag is not at the same layer of abstraction as the other flags above (such as number_of_messages). Is it sensible to specify both number_of_messages and max_publish_latency_test? I'm assuming 'yes', since the former specifies the stopping condition for an individual "run", whereas the latter specifies the stopping condition for a suite of such runs. So suggest breaking this up into two separate concepts: one which captures the aforementioned "suite", and one which captures a single iteration. IIUC, most existing flags apply to the "iteration". Then we each concept can be captured in its own class with its own flags.

load-test-framework/src/main/java/com/google/pubsub/clients/common/LoadTestRunner.java
```diff
@@ -48,21 +48,47 @@
...
+    final Semaphore outstandingTestLimiter =
+        new Semaphore(request.getMaxOutstandingRequests(), false);
+
+    final long toSleep = request.getStartTime().getSeconds() * 1000 - System.currentTimeMillis();
+    if (toSleep > 0) {
+      Thread.sleep(toSleep);
+    }
+
+    stopwatch.start();
+    while (shouldContinue(request)) {
```
mdietz94: This isn't about how many times it will be invoked, this is a cleanup focused around being agnostic to whether it is supposed to run for a duration or for a particular number of messages.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -164,6 +193,11 @@ private void run() {
       );
       Preconditions.checkArgument(
           broker != null || (kafkaPublisherCount == 0 && kafkaSubscriberCount == 0));
+      // If max publish latency is set, exactly one publisher type must be specified.
+      Preconditions.checkArgument(
+          maxPublishLatency == 0 || ((kafkaPublisherCount == 0 || cpsPublisherCount == 0) &&
+              kafkaPublisherCount + cpsPublisherCount > 0)
```
mdietz94: Done, didn't know there was a binary XOR operator, neat!

mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -141,6 +154,22 @@
           "message loss. If set less than 1, this flag is ignored."
   )
   private int numberOfMessages = 0;
+  @Parameter(
+      names = {"--max_publish_latency_test"},
+      description = "This sets the maximum latency, in this test we will continuously run load " +
```
mdietz94: Done.

mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -141,6 +154,22 @@
...
+          "tests with increasing request rates until we hit the provided latency. You must only " +
+          "provide a single type of publisher to use this test. This uses the 99% latency as the " +
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Done.

mdietz94:
PTAL

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -220,18 +273,78 @@ private void run() {
           log.info("===============================================");
         }
       }, 5, 5, TimeUnit.SECONDS);
-      // Wait for the load test to finish.
-      gceController.waitForClients();
+      Map<ClientType, Controller.LoadtestStats> statsMap;
+      AtomicDouble publishLatency = new AtomicDouble(0);
+      // This should be a number_of_messages test so that we know when the publisher is done.
+      // If the publisher has finished, check pubsub.googleapis.com/subscription/num_undelivered_messages.
+      // If greater than some threshold, like 1 pull, we fail and return the previous one.
```
amordkov: I don't understand this sentence. What is a 1-pull threshold?

load-test-framework/src/main/java/com/google/pubsub/clients/common/LoadTestRunner.java
```diff
@@ -94,33 +121,8 @@ public void run() {
...
-      rateLimiter.acquire();
-      executor.submit(client).addListener(outstandingTestLimiter::release, executor);
-    }
-    stopwatch.stop();
-    executor.shutdownNow();
-    finished.set(true);
-    finishedFuture.get();
-    log.info("Load test complete, shutting down.");
+    // Deadlock forever, since we do not want the server to stop.
+    Thread.currentThread().join();
```
amordkov: Why is this necessary now? Was it also it necessary before for handling Check requests after the test finished running?

load-test-framework/src/main/java/com/google/pubsub/clients/common/LoadTestRunner.java
amordkov: But you don't really need an instance of LoadTestRunner. The LoadTestRunner has no non-static member fields. It seems like all you need is a Function or lambda that you can schedule.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -156,20 +157,57 @@
...
   @Parameter(
       names = {"--max_subscriber_throughput_test"},
       description = "This test will continuously run load tests with greater publish request " +
           "rate until the subscriber can no longer keep up, and will let you know the maximum " +
           "throughput per subscribing client."
   )
   private boolean maxSubscriberThroughputTest = false;
-
+  @Parameter(
+      names = {"--max_subscriber_throughput_test_backlog"},
```
mdietz94: Done, messages.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -188,14 +226,17 @@ private void run() {
...
               || kafkaPublisherCount > 0
               || kafkaSubscriberCount > 0
-      );
+          , "You must set at least one type of client greater than 0.");
       Preconditions.checkArgument(
-          broker != null || (kafkaPublisherCount == 0 && kafkaSubscriberCount == 0));
+          broker != null || (kafkaPublisherCount == 0 && kafkaSubscriberCount == 0),
+          "If using Kafka you must provide the network address of your broker using the"
+              + "--broker flag.");
       // If max publish latency is set, exactly one publisher type must be specified.
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -220,18 +273,78 @@ private void run() {
           log.info("===============================================");
         }
       }, 5, 5, TimeUnit.SECONDS);
-      // Wait for the load test to finish.
-      gceController.waitForClients();
+      Map<ClientType, Controller.LoadtestStats> statsMap;
+      AtomicDouble publishLatency = new AtomicDouble(0);
+      // This should be a number_of_messages test so that we know when the publisher is done.
```
mdietz94: It could be either, this was a comment I wrote while I was thinking things through, meant to delete it. Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Removed, meant the size of a pull request.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -220,18 +273,78 @@ private void run() {
...
+          gceController.waitForPublisherClients();
+          ListTimeSeriesResponse response =
+              monitoring.projects().timeSeries().list("projects/" + project)
+                  .setFilter(
+                      "metric.type = \"pubsub.googleapis.com/subscription/num_undelivered_messages\"")
+                  .setIntervalStartTime(dateFormatter.format(startDate))
+                  .setIntervalEndTime(dateFormatter.format(new Date()))
+                  .execute();
+          // Get most recent point.
+          Point maxPoint = null;
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -220,18 +273,78 @@ private void run() {
...
+          for (TimeSeries timeSeries : response.getTimeSeries()) {
+            for (Point point : timeSeries.getPoints()) {
+              if (maxPoint == null ||
+                  dateFormatter.parse(point.getInterval().getStartTime()).after(
```
mdietz94: They are not guaranteed to be in order AFAIK.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -220,18 +273,78 @@ private void run() {
...
+                      dateFormatter.parse(maxPoint.getInterval().getStartTime()))) {
+                maxPoint = point;
+              }
+            }
+          }
+          if (maxPoint != null && maxPoint.getValue().getInt64Value() >
+              maxSubscriberThroughputTestBacklog) {
+            log.info("We accumulated a backlog during this test, refer to the last run " +
+                "for the maximum throughput capable before accumulating backlog.");
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -220,18 +273,78 @@ private void run() {
...
+            maxSubscriberThroughputTest = false;
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -131,11 +132,7 @@ void start() throws Throwable {
           .setSeconds(loadtestLengthSeconds).build());
     }
     switch (clientType) {
-      case CPS_GCLOUD_PUBLISHER:
-        requestBuilder.setPubsubOptions(PubsubOptions.newBuilder()
-            .setPublishBatchSize(cpsPublishBatchSize));
```
mdietz94: Turns out Kafka uses batching, so it's no longer PubSub specific.

load-test-framework/src/main/java/com/google/pubsub/clients/common/LoadTestRunner.java
mdietz94: Now that we don't want to stop after a single run, we need to have a thread sit doing nothing while the work is scheduled. Before we just waited on isFinished, but that doesn't make sense anymore since we'll run multiple tests.

mdietz94:
PTAL

load-test-framework/src/main/java/com/google/pubsub/clients/common/LoadTestRunner.java
amordkov: IMO, a constructor is special in the sense that it initializes and implicitly returns an object (which in this case we didn't even need) which is then used for something.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -142,6 +156,41 @@
...
+      names = {"--max_publish_latency_millis"},
+      description = "This sets the maximum latency in milliseconds, in this test we will" +
+          "continuously run load tests with increasing request rates until we hit the provided " +
+          "latency. You must only provide a single type of publisher to use this test. This " +
+          "uses the latency specified by max_publish_latency_percentile as the bound to check."
+  )
+  private int maxPublishLatencyMillis = 0;
+  @Parameter(
+      names = {"--max_publish_latency_percentile"},
+      description = "This sets the percentile to use when determining the latency for the max " +
```
amordkov: s/max publish latency test/max_publish_latency_test/

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -142,6 +156,41 @@
...
+      names = {"--max_publish_latency_test"},
+      description = "In this test we will continuously run load tests with increasing request " +
+          "rates until we hit max_publish_latency_millis. You must only provide a single type of " +
+          "publisher to use this test. This uses the latency specified by " +
+          "max_publish_latency_percentile as the bound to check."
+  )
+  private boolean maxPublishLatencyTest = false;
+  @Parameter(
+      names = {"--max_publish_latency_millis"},
+      description = "This sets the maximum latency in milliseconds, in this test we will" +
```
amordkov: This should just say "This is the maximum latency in milliseconds allowed before terminating the max_publish_latency_test.

load-test-framework/src/main/java/com/google/pubsub/clients/common/LoadTestRunner.java
mdietz94: Fair enough, it probably adds a little GC overhead anyway.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -165,16 +165,14 @@
   private boolean maxPublishLatencyTest = false;
   @Parameter(
       names = {"--max_publish_latency_millis"},
-      description = "This sets the maximum latency in milliseconds, in this test we will" +
-          "continuously run load tests with increasing request rates until we hit the provided " +
-          "latency. You must only provide a single type of publisher to use this test. This " +
-          "uses the latency specified by max_publish_latency_percentile as the bound to check."
+      description = "his is the maximum latency in milliseconds allowed before terminating the " +
```
amordkov: s/his/This/

PR #13, opened 2016-10-14
Load test stats output to Google Sheets

jrheizelman:
The stats generated by the test are packaged and outputted directly to a Google Sheets document, set by user by environment variable LOADTEST_SPREADSHEET_ID.

### DIFF
--- a/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
+++ b/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
@@ -204,8 +204,7 @@ class Driver {
       Client.numberOfMessages = numberOfMessages;
       GCEController gceController = GCEController.newGCEController(
           project, ImmutableMap.of("us-central1-a", clientParamsMap),
-          Executors.newScheduledThreadPool(500), cpsPublisherCount,
-          cpsSubscriberCount, kafkaPublisherCount, kafkaSubscriberCount);
+          Executors.newScheduledThreadPool(500));
       gceController.startClients();
 
       // Start a thread to poll and output results.
--- a/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
+++ b/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
@@ -301,6 +301,11 @@ class Driver {
     return new Sheets.Builder(transport, factory, credential).setApplicationName(APPLICATION_NAME).build();
   }
   
+  /* Publishes stats information to Google Sheets document. Format for sheet assumes the following
+   * column order: Publisher #; Subscriber #; Message size (B); Test length (s); # messages;
+   * Publish batch size; Subscribe pull size; Request rate; Max outstanding requests; 
+   * Throughput (MB/s); 50% (ms); 90% (ms); 99% (ms)
+   */
   private void sendToSheets(Sheets sheetService, List<String[]> stats) {
     List<List<Object>> cpsValues = new ArrayList<List<Object>>();
     List<List<Object>> kafkaValues = new ArrayList<List<Object>>();
@@ -353,7 +358,7 @@ class Driver {
       sheetService.spreadsheets().values().append(SPREADSHEET_ID, "Kafka", 
           new ValueRange().setValues(kafkaValues)).setValueInputOption("USER_ENTERED").execute();
     } catch (IOException e) {
-      e.printStackTrace();
+      log.error("Error publishing to spreadsheet: " + SPREADSHEET_ID);
     }
   }
 
--- a/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
+++ b/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
@@ -226,8 +226,8 @@ class Driver {
       
       if (spreadsheetId.length() > 0) {
         // Output results to common Google sheet
-        SheetsService service = new SheetsService(dataStoreDirectory);
-        service.sendToSheets(spreadsheetId, results, gceController);
+        SheetsService service = new SheetsService(dataStoreDirectory, gceController.getTypes());
+        service.sendToSheets(spreadsheetId, results);
       }
       gceController.shutdown(null);
       System.exit(0);
--- a/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
+++ b/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
@@ -31,35 +31,8 @@ import com.google.pubsub.flic.controllers.GCEController;
 import com.google.pubsub.flic.output.SheetsService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import com.google.api.client.auth.oauth2.Credential;
-import com.google.api.client.extensions.java6.auth.oauth2.AuthorizationCodeInstalledApp;
-import com.google.api.client.extensions.jetty.auth.oauth2.LocalServerReceiver;
-import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeFlow;
-import com.google.api.client.googleapis.auth.oauth2.GoogleClientSecrets;
-import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
-import com.google.api.client.http.HttpTransport;
-import com.google.api.client.json.JsonFactory;
-import com.google.api.client.json.jackson2.JacksonFactory;
-import com.google.api.client.util.store.FileDataStoreFactory;
-import com.google.api.services.sheets.v4.Sheets;
-import com.google.api.services.sheets.v4.SheetsScopes;
-import com.google.api.services.sheets.v4.model.AppendCellsRequest;
-import com.google.api.services.sheets.v4.model.BatchUpdateSpreadsheetRequest;
-import com.google.api.services.sheets.v4.model.CellData;
-import com.google.api.services.sheets.v4.model.ExtendedValue;
-import com.google.api.services.sheets.v4.model.Request;
-import com.google.api.services.sheets.v4.model.RowData;
-import com.google.api.services.sheets.v4.model.ValueRange;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.text.DecimalFormat;
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -172,7 +145,7 @@ class Driver {
       names = {"--spreadsheet_id"},
       description = "The id of the spreadsheet to which results are output."
   )
-  private String spreadsheetId;
+  private String spreadsheetId = "";
   @Parameter(
       names = {"--data_store_dir"},
       description = "The directory to store credentials for sheets output verification. Note: " +
@@ -229,6 +202,10 @@ class Driver {
       Client.maxOutstandingRequests = maxOutstandingRequests;
       Client.burnInTimeMillis = (Client.startTime.getSeconds() + burnInDurationSeconds) * 1000;
       Client.numberOfMessages = numberOfMessages;
+      Client.cpsPublisherCount = cpsPublisherCount;
+      Client.cpsSubscriberCount = cpsSubscriberCount;
+      Client.kafkaPublisherCount = kafkaPublisherCount;
+      Client.kafkaSubscriberCount = kafkaSubscriberCount;
       GCEController gceController = GCEController.newGCEController(
           project, ImmutableMap.of("us-central1-a", clientParamsMap),
           Executors.newScheduledThreadPool(500));
@@ -251,7 +228,7 @@ class Driver {
       Map<ClientType, Controller.LoadtestStats> results = gceController.getStatsForAllClientTypes();
       printStats(results);
       
-      if (spreadsheetId.length() < 0) {
+      if (spreadsheetId.length() > 0) {
         // Output results to common Google sheet
         SheetsService service = new SheetsService(dataStoreDirectory);
         service.sendToSheets(spreadsheetId, results);
--- a/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
+++ b/load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
@@ -30,9 +30,35 @@ import com.google.pubsub.flic.controllers.Controller;
 import com.google.pubsub.flic.controllers.GCEController;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import com.google.api.client.auth.oauth2.Credential;
+import com.google.api.client.extensions.java6.auth.oauth2.AuthorizationCodeInstalledApp;
+import com.google.api.client.extensions.jetty.auth.oauth2.LocalServerReceiver;
+import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeFlow;
+import com.google.api.client.googleapis.auth.oauth2.GoogleClientSecrets;
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.util.store.FileDataStoreFactory;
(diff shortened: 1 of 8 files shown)


load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -141,6 +167,11 @@
           "message loss. If set less than 1, this flag is ignored."
   )
   private int numberOfMessages = 0;
+  
+  private final String SPREADSHEET_ID = System.getenv("LOADTEST_SPREADSHEET_ID");
```
mdietz94: Make this a flag, instead of an environment variable.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -141,6 +167,11 @@
...
+  private final String APPLICATION_NAME = "loadtest-framework";
+  private final String DATA_STORE_DIR = 
+      System.getProperty("user.home") + "/.credentials/sheets.googleapis.com-loadtest-framework";
```
mdietz94: Also make this a flag.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -208,29 +239,122 @@ private void run() {
...
+      log.info("Results for " + row[0] + ":");
+      log.info("50%: " + row[1]);
+      log.info("99%: " + row[2]);
+      log.info("99.9%: " + row[3]);
+      // CPS Publishers report latency per batch message.
+      log.info("Average throughput: " + row[4] + " MB/s");
+    }
+  }
+  
+  private Sheets getSheetsService() throws Exception {
```
mdietz94: This belongs in a separate file. Why don't you make a com.google.pubsub.flic.output.Sheets and put everything in there, and have a public function that takes the List<String[]> stats and writes it out, this is too much for Driver

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -208,29 +239,122 @@ private void run() {
...
+    return new Sheets.Builder(transport, factory, credential).setApplicationName(APPLICATION_NAME).build();
+  }
+  
+  private void sendToSheets(Sheets sheetService, List<String[]> stats) {
+    List<List<Object>> cpsValues = new ArrayList<List<Object>>();
+    List<List<Object>> kafkaValues = new ArrayList<List<Object>>();
+    for (String[] row : stats) {
+      List<Object> valueRow = new ArrayList<Object>(13);
+      switch (row[0]) {
+        case "cps-gcloud-publisher":
```
mdietz94: This shouldn't operate on a List<String[]> this should operate on the entire Map, and you should use enum switching, not String for type safety.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
jrheizelman: That's what I had done originally, but it's a bit annoying putting in the value as a flag on every single run, and putting in our document as the default flag value would put the sheet id up publicly for all Github to see, which isn't too much of an issue I guess but I felt a bit weird about doing it.

mdietz94: Make the default empty, this should be something the user enables. You can use an alias so that the flag is provided by default for actually running it.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
jrheizelman: Fixed in SheetsService.java

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
jrheizelman: Fixed. I did keep this location set to be the default rather than empty string, because it still requires the feature being turned on by a user by setting spreadsheet_id. This being set to some default won't impact user experience unless they choose to use the feature.

jrheizelman:
Made the general changes in the recent commit to have better organization between driver/sheets output.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -226,7 +226,10 @@ public static GCEController newGCEController(
    */
   @Override
   public void shutdown(Throwable t) {
-    log.error("Shutting down: ", t);
+    if (t != null)
```
mdietz94: Always use curly brackets

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.LongStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SheetsService {
+  private final static Logger log = LoggerFactory.getLogger(SheetsService.class);
+  private Sheets service;
```
mdietz94: final

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -0,0 +1,125 @@
...
+  
+  /* Publishes stats information to Google Sheets document. Format for sheet assumes the following
+   * column order: Publisher #; Subscriber #; Message size (B); Test length (s); # messages;
+   * Publish batch size; Subscribe pull size; Request rate; Max outstanding requests; 
+   * Throughput (MB/s); 50% (ms); 90% (ms); 99% (ms)
+   */
+  public void sendToSheets(String sheetId, Map<ClientType, Controller.LoadtestStats> results) {
+    List<List<Object>> cpsValues = new ArrayList<List<Object>>(results.size());
+    List<List<Object>> kafkaValues = new ArrayList<List<Object>>(results.size());
+    results.forEach((type, stats) -> {
```
mdietz94: I think this should generalize, it would be better if it did not require knowledge ahead of time of the type of results present. That's more of a general nit, I'll accept this without those changes, but if possible it would be nicer.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -189,6 +202,10 @@ private void run() {
       Client.maxOutstandingRequests = maxOutstandingRequests;
       Client.burnInTimeMillis = (Client.startTime.getSeconds() + burnInDurationSeconds) * 1000;
       Client.numberOfMessages = numberOfMessages;
+      Client.cpsPublisherCount = cpsPublisherCount;
```
jrheizelman: I'm going to change this because I think the abstraction/design point is valid, but I have a question to clarify the O(2n) space usage. I made all the publisher/subscriber count variables static in the Client class, so my thinking there was it's stored once for the class per jvm. Is that right, or am I not understanding how java does that correctly?

mdietz94: That's correct, it isn't the static part I'm objecting to as much as the part it's part of the Client class. It could be part of the Controller / GCEController class.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
jrheizelman: Fixed

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
```diff
@@ -191,7 +204,8 @@ private void run() {
       Client.numberOfMessages = numberOfMessages;
       GCEController gceController = GCEController.newGCEController(
           project, ImmutableMap.of("us-central1-a", clientParamsMap),
-          Executors.newScheduledThreadPool(500));
+          Executors.newScheduledThreadPool(500), cpsPublisherCount,
```
mdietz94: The clientParamsMap already has all these counts. It makes more sense to just iterate over the map to build your spreadsheet.

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
jrheizelman: I think it would be a good bit harder to adapt to different spreadsheet styles, and once you require that knowledge, I don't think the usage gains from not needing to know result types actually helps you that much. I'll keep this in mind as a todo, but I don't think it's worth the time right now.

load-test-framework/src/main/java/com/google/pubsub/flic/Driver.java
jrheizelman: Fixed

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -67,24 +67,41 @@
...
+            cpsSubscriberCount += count;
+            break;
+          case KAFKA_PUBLISHER:
+            kafkaPublisherCount += count;
+            break;
+          case KAFKA_SUBSCRIBER:
+            kafkaSubscriberCount += count;
+            break;
+        }
+      });
```
mdietz94: How about:
`cpsPublisherCount = types.values().forEach(paramsMap -> paramsMap.keySet().stream().map(p -> p.clientType).filter(t -> t == CPS_GCLOUD_PUBLISHER).count()` instead of this loop

mdietz94: I just updated the comment, the old version wouldn't actually work quite nicely, I think this is a little cleaner.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/GCEController.java
```diff
@@ -431,4 +449,28 @@ private InstanceTemplate defaultInstanceTemplate(String type) {
             .setServiceAccounts(Collections.singletonList(new ServiceAccount().setScopes(
                 Collections.singletonList("https://www.googleapis.com/auth/cloud-platform")))));
   }
+
+  /**
```
mdietz94: I don't like this, again this scales O(n) with new types. Please just expose the map. Now I'm starting to feel like you should really just expose the original types map passed in, and do processing on it in the Sheets file.

jrheizelman: I moved the code to the sheets service for now. This works for now, but when we get more clients, I'll add the changes we discussed to read client type, check cps/kakfa and publish/subscribe, before adding to the spreadsheet.

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -0,0 +1,140 @@
...
+      cpsPublisherCount += countMap.get(ClientType.CPS_GCLOUD_PUBLISHER);
+      cpsSubscriberCount += countMap.get(ClientType.CPS_GCLOUD_SUBSCRIBER);
+      kafkaPublisherCount += countMap.get(ClientType.KAFKA_PUBLISHER);
+      kafkaSubscriberCount += countMap.get(ClientType.KAFKA_PUBLISHER);
+    });
+    results.forEach((type, stats) -> {
+      List<Object> valueRow = new ArrayList<Object>(13);
+      switch (type) {
+        case CPS_GCLOUD_PUBLISHER:
+          if (cpsPublisherCount == 0) return;
```
mdietz94: Please use curly brackets.

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -0,0 +1,140 @@
...
+          cpsValues.add(valueRow);
+          break;
+        case CPS_GCLOUD_SUBSCRIBER:
+          if (cpsSubscriberCount == 0) return;
+          valueRow.add(0);
+          valueRow.add(cpsSubscriberCount);
+          cpsValues.add(valueRow);
+          break;
+        case KAFKA_PUBLISHER:
+          if (kafkaPublisherCount == 0) return;
```
mdietz94: Ditto.

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -0,0 +1,140 @@
...
+          valueRow.add(kafkaPublisherCount);
+          valueRow.add(0);
+          kafkaValues.add(valueRow);
+          break;
+        case KAFKA_SUBSCRIBER:
+          if (kafkaSubscriberCount == 0) return;
```
mdietz94: Ditto

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
jrheizelman: Fixed all

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -0,0 +1,140 @@
...
+      List<Object> valueRow = new ArrayList<Object>(13);
+      switch (type) {
+        case CPS_GCLOUD_PUBLISHER:
+          if (cpsPublisherCount == 0) return;
+          valueRow.add(cpsPublisherCount);
+          valueRow.add(0);
+          cpsValues.add(valueRow);
+          break;
+        case CPS_GCLOUD_SUBSCRIBER:
+          if (cpsSubscriberCount == 0) return;
```
jrheizelman: Fixed

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
jrheizelman: Fixed

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -0,0 +1,140 @@
...
+    service = new Sheets.Builder(transport, factory, credential)
+        .setApplicationName(APPLICATION_NAME).build();
+  }
+  
+  /* Publishes stats information to Google Sheets document. Format for sheet assumes the following
+   * column order: Publisher #; Subscriber #; Message size (B); Test length (s); # messages;
+   * Publish batch size; Subscribe pull size; Request rate; Max outstanding requests; 
+   * Throughput (MB/s); 50% (ms); 90% (ms); 99% (ms)
+   */
+  public void sendToSheets(String sheetId, Map<ClientType, Controller.LoadtestStats> results,
```
mdietz94: Separate into a function that outputs the values, and a function that writes to Sheets. Then please write a unit test for the former. This code is going to get changed with every new type, and it would be good to catch errors in our tests.

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -0,0 +1,150 @@
...
+      cpsPublisherCount += countMap.get(ClientType.CPS_GCLOUD_PUBLISHER);
+      cpsSubscriberCount += countMap.get(ClientType.CPS_GCLOUD_SUBSCRIBER);
+      kafkaPublisherCount += countMap.get(ClientType.KAFKA_PUBLISHER);
+      kafkaSubscriberCount += countMap.get(ClientType.KAFKA_PUBLISHER);
+    });
+    results.forEach((type, stats) -> {
+      List<Object> valueRow = new ArrayList<Object>(13);
+      switch (type) {
+        case CPS_GCLOUD_PUBLISHER:
+          if (cpsPublisherCount == 0) { return; }
```
mdietz94: Also newlines with the brackets.
```
if (cpsPublisherCount == 0) {
  return;
}
```

load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -147,4 +161,32 @@ public void sendToSheets(String sheetId, Map<ClientType, Controller.LoadtestStat
     out.add(kafkaValues);
     return out;
   }
+
+  /**
+   * @return the cpsPublisherCount
+   */
+  public int getCpsPublisherCount() {
```
jrheizelman: Fixed

jrheizelman:
Now the unit tests should work for all cps and kafka clients we add to the ClientType enum. If we add one and don't either add it to the switch/service logic and/or get the changes done to generalize it, then they'll fail.

mdietz94:
Great, LGTM

PR #14, opened 2016-10-17
Remote Code Support

mdietz94:
Support for RemoteTasks that are in other languages, with a working PythonPublisher example.


load-test-framework/python_src/clients/cps_publisher_task.py
```diff
@@ -0,0 +1,60 @@
...
+        return response
+
+
+if __name__ == "__main__":
+    server = grpc.server(futures.ThreadPoolExecutor(max_workers=50))
+    loadtest_pb2.add_LoadtestServicer_to_server(LoadtestServicer(), server)
+    server.add_insecure_port('localhost:6000')
+    server.start()
+    while True:
+        time.sleep(3600)
```
amordkov: Why sleep?

load-test-framework/python_src/clients/cps_publisher_task.py
```diff
@@ -0,0 +1,60 @@
...
+from google.cloud import pubsub
+
+import loadtest_pb2
+
+
+class LoadtestServicer(loadtest_pb2.LoadtestServicer):
+    """Provides methods that implement functionality of load test server."""
+
+    def __init__(self):
+        self.pubsub_client = pubsub.Client()
```
amordkov: You don't really need this as a member variable.

load-test-framework/python_src/clients/cps_publisher_task.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def Start(self, request, context):
+        self.message_size = request.message_size
+        self.batch_size = request.pubsub_options.publish_batch_size
+        self.batch = self.pubsub_client.topic(request.topic).batch()
+        return loadtest_pb2.StartResponse()
+
+    def Check(self, request, context):
+        if not self.batch:
+            return loadtest_pb2.CheckResponse()
```
amordkov: Seems like this should throw an error instead. It is wrong to call Check before calling Start.

load-test-framework/python_src/clients/cps_publisher_task.py
```diff
@@ -0,0 +1,60 @@
...
+        start = time.clock()
+        for i in range(0, self.batch_size):
+            self.batch.publish(("A" * self.message_size).encode(), sendTime=str(int(start * 1000)))
```
amordkov: What is this sendTime parameter? How is it used?

load-test-framework/src/main/java/com/google/pubsub/clients/gcloud/CPSPublisherTask.java
```diff
@@ -70,7 +70,7 @@ public void run() {
       pubSub.publish(topic, messages);
       stopwatch.stop();
       numberOfMessages.addAndGet(batchSize);
-      metricsHandler.recordLatency(stopwatch.elapsed(TimeUnit.MILLISECONDS));
```
amordkov: Was this just incorrect before?

load-test-framework/src/main/java/com/google/pubsub/clients/remote/RemoteTask.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+
+
+/**
+ * Runs a task that executes command on the remote server.
+ */
+class RemoteTask extends Task {
```
amordkov: Given that the loadtest framework already consists of code that runs locally (Driver) and code that runs remotely ("Client", which runs in VMs), the term remote here is confusing. I think this should be called AdapterTask -- it adapts various tasks (in other languages) to work with the primary Java framework. The directory should also be renamed "remote" to "adapter", although I don't actually see the need for this to be in its own sub-package.

load-test-framework/src/main/java/com/google/pubsub/clients/remote/RemoteTask.java
```diff
@@ -0,0 +1,68 @@
...
+      stub.start(request);
+    } catch (Throwable t) {
+      log.error("Unable to start server.", t);
+      System.exit(1);
+    }
+  }
+
+  public static void main(String[] args) throws Exception {
+    LoadTestRunner.run(request -> {
+      boolean publisher = Arrays.stream(args).anyMatch(x -> x.startsWith("publish"));
```
amordkov: This seems super brittle.

load-test-framework/src/main/java/com/google/pubsub/clients/remote/RemoteTask.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Runs a task that executes command on the remote server.
+ */
+class RemoteTask extends Task {
+  private static final Logger log = LoggerFactory.getLogger(RemoteTask.class);
+  private LoadtestGrpc.LoadtestBlockingStub stub;
+
+  private RemoteTask(LoadtestProto.StartRequest request, MetricsHandler.MetricName metricName) {
+    super(request.getProject(), "remote", metricName);
```
amordkov: s/remote/adapter/g

load-test-framework/src/main/java/com/google/pubsub/clients/remote/RemoteTask.java
```diff
@@ -0,0 +1,68 @@
...
+  @Override
+  public void run() {
+    try {
+      LoadtestProto.CheckResponse response =
+          stub.check(LoadtestProto.CheckRequest.getDefaultInstance());
+      numberOfMessages.addAndGet(response.getBucketValuesCount());
+      response.getBucketValuesList().forEach(metricsHandler::recordLatency);
+    } catch (Throwable t) {
+      log.error("Error running command on remote task.", t);
+    }
```
amordkov: This is super confusing, since its conflating number of points with number of buckets.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -81,6 +81,8 @@ static String getTopicSuffix(ClientType clientType) {
       case CPS_GCLOUD_PUBLISHER:
       case CPS_GCLOUD_SUBSCRIBER:
         return "gcloud";
+      case CPS_GCLOUD_PYTHON_PUBLISHER:
+        return "python";
```
amordkov: This doesn't look right. Until now, these strings reflected the service used for publishing, not the language of the client library. Or are they supposed to reflect the client library type?

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -81,6 +81,8 @@ static String getTopicSuffix(ClientType clientType) {
       case CPS_GCLOUD_PUBLISHER:
```
amordkov: Should this be CPS_GCLOUD_JAVA_PUBLISHER now? Same goes for CPS_GCLOUD_SUBSCRIBER. Actually, what is the 'GCLOUD' referring to? The client library? If so, does CPS_GCLOUD_PYTHON_PUBLISHER even make sense?

load-test-framework/python_src/clients/cps_publisher_task.py
mdietz94: I do if I don't want to recreate it on every check call. This creates stubs internally.

load-test-framework/python_src/clients/cps_publisher_task.py
mdietz94: It is used by the Java client as well, the subscriber uses it to calculate end to end latency.

load-test-framework/python_src/clients/cps_publisher_task.py
mdietz94: Because otherwise the program will exit, according to what I've read this is the recommended way of waiting without polling.

load-test-framework/src/main/java/com/google/pubsub/clients/gcloud/CPSPublisherTask.java
mdietz94: No, before I was multiplying by the batch size later, but it's definitely better if every Task does their exporting the same.

load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
mdietz94: You are correct, thanks, Done.

load-test-framework/src/main/java/com/google/pubsub/clients/remote/RemoteTask.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/clients/remote/RemoteTask.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/clients/remote/RemoteTask.java
mdietz94: Fixed to actually use types.

load-test-framework/python_src/clients/cps_publisher_task.py
mdietz94: Fair point, I just removed the check so it will return a NullPointerError. In reality this is sort of impossible, since we don't even have the stub to call Check on until Start comes back.

load-test-framework/src/main/java/com/google/pubsub/clients/remote/RemoteTask.java
mdietz94: Done, new service called Adapter with Start/Execute commands.

mdietz94:
PTAL

load-test-framework/src/main/java/com/google/pubsub/clients/adapter/AdapterSubscriberTask.java
```diff
@@ -0,0 +1,34 @@
...
+package com.google.pubsub.clients.adapter;
+
+import com.google.pubsub.clients.common.LoadTestRunner;
+import com.google.pubsub.clients.common.MetricsHandler;
+import com.google.pubsub.flic.common.LoadtestProto;
+
+/**
+ * Runs a task that publishes on the adapter server.
+ */
+class AdapterSubscriberTask extends AdapterTask {
```
amordkov: SubscriberAdapterTask

load-test-framework/src/main/java/com/google/pubsub/clients/adapter/AdapterSubscriberTask.java
```diff
@@ -0,0 +1,34 @@
...
+//
+////////////////////////////////////////////////////////////////////////////////
+package com.google.pubsub.clients.adapter;
+
+import com.google.pubsub.clients.common.LoadTestRunner;
+import com.google.pubsub.clients.common.MetricsHandler;
+import com.google.pubsub.flic.common.LoadtestProto;
+
+/**
+ * Runs a task that publishes on the adapter server.
```
amordkov: subscribes on?

load-test-framework/src/main/java/com/google/pubsub/clients/adapter/AdapterPublisherTask.java
```diff
@@ -0,0 +1,34 @@
...
+//
+////////////////////////////////////////////////////////////////////////////////
+package com.google.pubsub.clients.adapter;
+
+import com.google.pubsub.clients.common.LoadTestRunner;
+import com.google.pubsub.clients.common.MetricsHandler;
+import com.google.pubsub.flic.common.LoadtestProto;
+
+/**
+ * Runs a task that publishes on the adapter server.
```
amordkov: The "adapter" is the java part, IIUC. Is the "adapter server" referring to the task in the foreign language? Per recommendation below, suggest s/adapter server/worker/.

Same in AdapterSubscriberTask.

load-test-framework/src/main/java/com/google/pubsub/clients/adapter/AdapterTask.java
```diff
@@ -0,0 +1,57 @@
...
+
+
+/**
+ * Runs a task that executes command on the adapter server.
+ */
+class AdapterTask extends Task {
+  private static final Logger log = LoggerFactory.getLogger(AdapterTask.class);
+  private AdapterGrpc.AdapterBlockingStub stub;
+
+  AdapterTask(LoadtestProto.StartRequest request, MetricsHandler.MetricName metricName) {
```
amordkov: line length?

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -16,6 +16,14 @@ service Loadtest {
   rpc Check(CheckRequest) returns (CheckResponse);
 }
 
+service Adapter {
+  // Starts an adapter
+  rpc Start (StartRequest) returns (StartResponse);
+
+  // Executes a command on the adapter, returning the latencies to do so.
```
amordkov: s/to do so/of the operations/

Also, if there is a single command, how come there are multiple latencies (clarify in the comment).

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -16,6 +16,14 @@ service Loadtest {
   rpc Check(CheckRequest) returns (CheckResponse);
 }
 
+service Adapter {
```
amordkov: So this service itself is not an "adapter" -- the adapter part is the AdapterTask which runs in Java makes an rpc to this service.
The service here is more of a 'slave' or 'worker'. Suggest calling this LoadtestSlave, LoadtestWorker, or just Worker. I prefer LoadtestWorker, but any of those is fine.

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -83,4 +91,12 @@ message CheckResponse {
 
   // True if the load test has finished running.
   bool is_finished = 3;
+}
+
+message ExecuteRequest {
+}
+
+message ExecuteResponse {
+  // Latencies to reportfor the given action
```
amordkov: "Latencies of the completed operations"

load-test-framework/src/main/java/com/google/pubsub/clients/adapter/AdapterPublisherTask.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+class AdapterPublisherTask extends AdapterTask {
```
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/clients/adapter/AdapterSubscriberTask.java
mdietz94: Done.

load-test-framework/src/main/java/com/google/pubsub/clients/adapter/AdapterTask.java
mdietz94: This line is less than 100 characters, the Java style guide limit.

load-test-framework/src/main/proto/loadtest.proto
mdietz94: Done.

load-test-framework/src/main/proto/loadtest.proto
mdietz94: Done.

load-test-framework/src/main/proto/loadtest.proto
mdietz94: Done.

mdietz94:
PTAL

load-test-framework/src/main/proto/loadtest.proto
```diff
@@ -16,8 +16,18 @@ service Loadtest {
...
+  // different end to end latencies) a single command can have multiple latencies returned.
+  rpc Execute (ExecuteRequest) returns (ExecuteResponse);
+}
+
 message StartRequest {
-  // The GCP project. This must be set even for Kafka, as we use it to export metrics to Stackdriver.
+  // The GCP project. This must be set even for Kafka, as we use it to export metrics.
   string project = 1;
 
   // The Pub/Sub or Kafka topic name.
```
amordkov: rogue comment?

mdietz94: Over the length limit.

PR #16, opened 2016-10-20
Made imports explicit

mdietz94:


jrheizelman:
Assuming this compiles and the tests pass. LGTM

PR #21, opened 2016-10-26
Cps throughput fix

jrheizelman:
Fixes an issue where the CPS gcloud publisher's latency buckets only recording 1 record for each batch, so the throughput numbers were off by a multiple of that size. 

Also adds a unit test to test equivalence between old recordLatency run n times and the batch recordLatency with batch size n.


load-test-framework/src/main/java/com/google/pubsub/clients/common/MetricsHandler.java
```diff
@@ -177,11 +177,15 @@ private void createMetrics() {
       log.info("Metrics already exist.");
     }
   }
-
+
```
mdietz94: Remove weird whitespace.

load-test-framework/src/main/java/com/google/pubsub/flic/common/LatencyDistribution.java
```diff
@@ -124,7 +124,7 @@ public synchronized LatencyDistribution copy() {
   public List<Long> getBucketValuesAsList() {
     return Arrays.asList(ArrayUtils.toObject(bucketValues));
   }
-
+
```
mdietz94: Weird whitespace.

load-test-framework/src/main/java/com/google/pubsub/flic/common/LatencyDistribution.java
```diff
@@ -146,4 +146,26 @@ public void recordLatency(long latencyMs) {
       bucketValues[LATENCY_BUCKETS.length - 1]++;
     }
   }
+
+  public void recordLatency(long latencyMs, int addCount) {
```
mdietz94: May want to make this recordLatencyBatch, and have a recordLatency that just does 1, since we use that a lot and it may look nicer.  Also maybe batchSize instead of addCount. We prefer not to use overloading whenever possible.

load-test-framework/src/main/java/com/google/pubsub/flic/common/LatencyDistribution.java
jrheizelman: Removed

load-test-framework/src/main/java/com/google/pubsub/flic/common/LatencyDistribution.java
jrheizelman: Renamed

load-test-framework/src/main/java/com/google/pubsub/clients/common/MetricsHandler.java
jrheizelman: Removed

PR #25, opened 2016-11-07
Script fixes

jrheizelman:
Installing java 8 is failing on all runs of framework due to a change in ubuntu's repository for the jdk. To prevent this from happening again, added apt-get update to the startup scripts.

Also moved creation of Client.startime higher up in the Driver class to avoid null pointer at line 267


mdietz94:
I changed your bugfix, what you did would have broken the new changes, the correct solution is to move line 267 down, but thanks for the fix.

jrheizelman:
Ok cool. I didn't spend that long looking at it, so thanks for checking. I'll merge when the tests finish.

PR #26, opened 2016-11-07
Use STRUCT schema for source connector when the CPS message contains attributes

kamalaboulhosn:


kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
```diff
@@ -122,23 +126,62 @@ public void start(Map<String, String> props) {
...
-        // data is coming from CPS so it must be of that type.
-        SourceRecord record =
+        String key = messageAttributes.get(kafkaMessageKeyAttribute);
+        ByteString messageData = message.getData();
+        ByteBuffer messageBytes = messageData.asReadOnlyByteBuffer();
+
+        boolean hasAttributes =
+            messageAttributes.size() > 1 ||
+            (messageAttributes.size() > 0 &&
+             !messageAttributes.containsKey(kafkaMessageKeyAttribute));
```
jrheizelman: Just a nit, but wouldn't && key == null be more efficient here?

kafka-connector/src/main/java/com/google/pubsub/kafka/common/ConnectorUtils.java
```diff
@@ -43,6 +43,7 @@
   public static final String CPS_MESSAGE_KEY_ATTRIBUTE = "key";
   public static final String CPS_MESSAGE_PARTITION_ATTRIBUTE = "kafka_partition";
   public static final String CPS_MESSAGE_KAFKA_TOPIC_ATTRIBUTE = "kafka_topic";
+  public static final String CPS_MESSAGE_KAFKA_STRUCT_ATTRIBUTE = "message";
```
jrheizelman: The naming scheme here is a bit inconsistent. For the above declarations, CPS_MESSAGE_***_ATTRIBUTE means the attribute name in the CPS message, but this means the attribute name in the Kafka struct for the CPS message. I think it'd be better if whichever side's attribute the string is the key for is always either first or last.

kamalaboulhosn:
We might consider getting rid of sending partition and topic info through on the sink connector. I'm not sure how much value they add. As far as those properties resulting in a struct if it comes back through the source connector, we could consider removing those properties explicitly if that does happen. But I think it makes sense for us to do the sink side first and then make the call. What do you think?

jrheizelman:
I don't think the partition or topic info adds too much value either. It would be the partition the message came from, which I don't think means that much once you're out of Kafka. And if you republish back to Kafka using the source it would be on a different topic and have different partition info. 

In that case, we can plan on removing them when one of us redoes the sink and not worry about them coming in attributes here.

PR #28, opened 2016-11-14
Grpc client added, Kafka publisher, and stats changes

jrheizelman:
Added a grpc client using the api currently available in alpha. Additionally, made changes to the sheets output formatting to include client information and better data output for what we've found we need.


jrheizelman:
Lots of changes in this PR, here's a quick, non-exhaustive list:
- Sheets output formatting changes to include better qps data
- Stats tracking on time wasted between batch publishes to help diagnose low qps numbers
- Addition of grpc publisher and subscriber
- Changes to the way Kafka publisher handles callbacks, batching based on Confluent recommendations
- CLI print output changed to reduce amount of unnecessary output

denyska:
@jrheizelman do we still need this PR. If no then lets close.

PR #29, opened 2016-11-15
Sink connector

jrheizelman:
Added a sink connector, supporting all primitive schema types, and arrays, structs, and maps with the following conditions:

Value types for an array cannot be struct or map.
Map must be String to String.
Struct value types must be Strings, with the exception of the value for the field to be placed in the message body, which can be any type other than struct or map. That field name is given by messageBodyName.


jrheizelman:
I realized after going home that I forgot to add in message size tracking for batching, so this isn't yet finished. Will add that now, but if you're reviewing it, ignore that deficiency.

jrheizelman:
Ok, issue is fixed now, so it will track the size of all attributes added from maps/structs.

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -165,6 +157,102 @@ public void put(Collection<SinkRecord> sinkRecords) {
...
+        return ByteString.copyFrom(doubleBuf);
+      case BOOLEAN:
+        byte bool = (byte)((Boolean) value?1:0);
+        byte[] boolArr = {bool};
+        return ByteString.copyFrom(boolArr);
+      case STRING:
+        String str = (String) value;
+        return ByteString.copyFromUtf8(str);
+      case BYTES:
+        return (ByteString) value;
```
kamalaboulhosn: Can we be sure this is a ByteString as opposed to, say, a byte[] or ByteBuffer?

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -165,6 +157,102 @@ public void put(Collection<SinkRecord> sinkRecords) {
...
+        for (Field f : fields) {
+          if (f.name().equals(messageBodyName)) {
+            Schema bodySchema = f.schema();
+            try {
+              msgBody = handleValue(bodySchema, struct.get(f), null);
+            } catch (NullPointerException e) { // Caused by accessing null attributes value
+              throw new DataException("Struct message body does not support Map or Struct types.");
+            }
+          } else {
+            if (f.schema().type() != Type.STRING) {
```
kamalaboulhosn: We could instead do toString() on the value.

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -165,6 +157,102 @@ public void put(Collection<SinkRecord> sinkRecords) {
...
+            attributes.put(f.name(), val);
+          }
+        }
+        if (msgBody != null) {
+          return msgBody;
+        } else {
+          return ByteString.EMPTY;
+        }
+      case MAP:
+        if (schema.keySchema().type() != Schema.Type.STRING ||
```
kamalaboulhosn: Could just do toString here, too.

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -165,6 +157,102 @@ public void put(Collection<SinkRecord> sinkRecords) {
...
+      case ARRAY:
+        Schema.Type arrType = schema.valueSchema().type();
+        if (arrType == Type.MAP || arrType == Type.STRUCT) {
+          throw new DataException("Array type does not support Map or Struct types.");
+        }
+        ByteString out = ByteString.EMPTY;
+        Object[] objArr = (Object[]) value;
+        for (Object o : objArr) {
+          out = out.concat(handleValue(schema.valueSchema(), o, null));
+        }
```
kamalaboulhosn: Should this return out?

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -110,27 +114,15 @@ public void put(Collection<SinkRecord> sinkRecords) {
     log.debug("Received " + sinkRecords.size() + " messages to send to CPS.");
     PubsubMessage.Builder builder = PubsubMessage.newBuilder();
     for (SinkRecord record : sinkRecords) {
-      if (record.valueSchema().type() != Schema.Type.BYTES
-          || !record.valueSchema().name().equals(ConnectorUtils.SCHEMA_NAME)) {
-        throw new DataException("Unexpected record of type " + record.valueSchema());
-      }
-      log.trace("Received record: " + record.toString());
```
kamalaboulhosn: Let's keep the trace statement; it can be helpful for debugging.

jrheizelman: Done

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
jrheizelman: Done

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
jrheizelman: Done.

jrheizelman:
Fixed those issues. I'm not totally sure what the BYTES schema type means for the value class in all possible cases (previously, it casted straight to ByteString), but hopefully byte[], ByteString, and ByteBuffer covers all the possible types it could have.

PR #48, opened 2017-02-21
update gcloud Tasks with high perf client

pongad:
I verified that tests are working. However, we should not submit this, until `google-cloud-pubsub` publishes `0.9.3-alpha` (without the SNAPSHOT). Otherwise all builds will fail.

### DIFF
--- a/load-test-framework/pom.xml
+++ b/load-test-framework/pom.xml
@@ -128,7 +128,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsub</artifactId>
-      <version>0.9.3-alpha-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


pongad:
This should be working now

mdietz94:
Just to confirm, is this safe for me to merge?

pongad:
@mdietz94 I think so, maybe wait for Travis to make sure?

mdietz94:
Travis tests are broken, they were written by a former intern and are
unreliable. If you've run this code through the framework and it's worked
that's fine, the code LGTM

pongad:
@mdietz94 Ah I see. The code does work when I tried it. Though Travis is failing, it also verified that the new package is up and compiling. (It fails while testing but compiled OK).  I think we should be good.

PR #51, opened 2017-03-04
Client code has been modified to be compatible with jdk1.7.

mprokhorenko:
- _AbstractSubscriberConnection#addOutstadingAckHandlers_ java 1.8 construction (..putIfAbsent) has been replaced with [V v = map.get(key); if (v == null) v = map.put(key, value);](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#putIfAbsent-K-V-)
- _SubscriberImplTest#testReceiverError_NacksMessage_ looks like unstable  and failed in most travis builds.  _Thread.sleep(200)_ helped, better solution should be provided.
- _.travis.yaml_ scripts has been updated with different jdk versions, all projects builds with jdk 1.7/1.8 except 'load-test-framework'

### DIFF
--- a/client/src/main/java/com/google/cloud/pubsub/AbstractSubscriberConnection.java
+++ b/client/src/main/java/com/google/cloud/pubsub/AbstractSubscriberConnection.java
@@ -310,7 +310,10 @@ abstract class AbstractSubscriberConnection extends AbstractService {
 
   private void addOutstadingAckHandlers(
       ExpirationInfo expiration, final List<AckHandler> ackHandlers) {
-    outstandingAckHandlers.putIfAbsent(expiration, new ArrayList<AckHandler>(ackHandlers.size()));
+    if (outstandingAckHandlers.get(expiration) == null)
+    {
+      outstandingAckHandlers.put(expiration, new ArrayList<AckHandler>(ackHandlers.size()));
+    }
     outstandingAckHandlers.get(expiration).addAll(ackHandlers);
   }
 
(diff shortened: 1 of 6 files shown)


client/src/main/java/com/google/cloud/pubsub/AbstractSubscriberConnection.java
```diff
@@ -310,7 +310,10 @@ public void run() {
 
   private void addOutstadingAckHandlers(
       ExpirationInfo expiration, final List<AckHandler> ackHandlers) {
-    outstandingAckHandlers.putIfAbsent(expiration, new ArrayList<AckHandler>(ackHandlers.size()));
```
mprokhorenko: good question, based on [java doc](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#putIfAbsent-K-V-), default imlementation doesn't have such guarantees. 

@davidtorresv do you have any idea or preferences on this change?

davidtorres: It is expected to be atomic, you might want to check the library in https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-pubsub

It is based on this code and will replace it.

mprokhorenko: @denyska, @davidtorres all modification related to _outstandingAckHandlers_ state made in synchronized block by itself, means we are quite atomic here.

Please approve PR if you don't have other questions.

PR #53, opened 2017-03-08
make LoadTestRunner back off

pongad:
This commit makes LoadTestRunner block when the task queue of its
executor is full.

Previously the LTR would discard the old task and immediate queue a new
one. Depending on how CPU resource is shared between threads and
processes, this could starve the process under test.

cc @mdietz94 @jba @garrettjonesgoogle


mdietz94:
LGTM

PR #55, opened 2017-03-13
add go publisher

pongad:
This PR also fixes a race in AdapterTask.
Previously it issues a StartRequest to the worker process without
waiting for the RPC to complete.
Consequently, the load test might happen before the worker process
could process the StartRequest.
This is fixed by making AdapterTask wait for the request to return.

The Go support is incomplete.
The program expects the binary to be available at gs://BUCKET/loadtest-go,
but does not automatically upload it there.
If possible, I'd like to land this automation in a separate PR to unblock my work
on load testing Go subscriber implementation.

cc @jba

### DIFF
--- a/load-test-framework/src/main/resources/gce/cps-gcloud-go-publisher_startup_script.sh
+++ b/load-test-framework/src/main/resources/gce/cps-gcloud-go-publisher_startup_script.sh
@@ -1,0 +1,38 @@
+#/bin/bash
+
+#######################################
+# Query GCE for a provided metadata field.
+# See https://developers.google.com/compute/docs/metadata
+# Globals:
+#   None
+# Arguments:
+#   $1: The path to the metadata field to retrieve
+# Returns:
+#   The value stored at the metadata field
+#######################################
+function metadata() {
+  curl --silent --show-error --header 'Metadata-Flavor: Google' \
+    "http://metadata/computeMetadata/v1/${1}";
+}
+
+readonly TMP="$(mktemp -d)"
+readonly BUCKET=$(metadata instance/attributes/bucket)
+
+[[ "${TMP}" != "" ]] || error mktemp failed
+
+# Download the loadtest binary to this machine and install Java 8.
+/usr/bin/apt-get update
+/usr/bin/apt-get install -y openjdk-8-jre-headless &
+/usr/bin/gsutil cp "gs://${BUCKET}/driver.jar" "${TMP}" &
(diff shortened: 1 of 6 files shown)


load-test-framework/src/main/resources/gce/cps-gcloud-go-publisher_startup_script.sh
```diff
@@ -0,0 +1,38 @@
+#/bin/bash
```
jba: #!

?

load-test-framework/src/main/resources/gce/cps-gcloud-go-publisher_startup_script.sh
```diff
@@ -0,0 +1,38 @@
...
+#######################################
+function metadata() {
+  curl --silent --show-error --header 'Metadata-Flavor: Google' \
+    "http://metadata/computeMetadata/v1/${1}";
+}
+
+readonly TMP="$(mktemp -d)"
+readonly BUCKET=$(metadata instance/attributes/bucket)
+
+[[ "${TMP}" != "" ]] || error mktemp failed
```
jba: I'm not familiar with the "error" command, and it's not installed on my workstation. What does it do?

pongad:
@mdietz94 Working in dev branch is OK with me. I'd like to close out comments by @jba before closing the issue if you don't mind.

load-test-framework/src/main/resources/gce/cps-gcloud-go-publisher_startup_script.sh
pongad: oh dear, thank you for catching this. I'm not sure how it worked all this time

load-test-framework/src/main/resources/gce/cps-gcloud-go-publisher_startup_script.sh
pongad: The `[[ "${TMP}" != "" ]] || error mktemp failed` is in all the startup scripts. I actually thought it was a shell builtin I didn't know about. After further testing, it does look like the command doesn't exist.

@mdietz94 do you know what's going on?

mdietz94: This was copied from a different loadtest script we had written. It is not a command, it just causes the script to exit since the error command doesn't exist. We can change this to something else, but from the load test it will just look like a byzantine error regardless.

PR #57, opened 2017-03-17
Fixing POM file to get rid of maven warnings

urshunkeler:
Figuring out how to work with github

### DIFF
--- a/jms-light/pom.xml
+++ b/jms-light/pom.xml
@@ -16,6 +16,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.6.1</version>
         <configuration>
           <source>${java.version}</source>
           <target>${java.version}</target>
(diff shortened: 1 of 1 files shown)


denyska:
@urshunkeler  do you know why Travis is failing ?

urshunkeler:
Hi Denys,

Thanks for your message.

The pull request to fix the POM file to get rid of maven warnings fails
only on the Oracle JDK 8 configuration and for the target
kafka-connector/pom.xml (the OpenJDK configurations pass). The target for
jms-light/pom.xml actually passed all tests. I do not understand the
reasons for the kafka-connector tests to fail. I have not touched anything
outside jms-light.

The pull request to test for compliance with the Google coding style fails
only on the Oracle JDK 8 configuration and for the target client/pom.xml
(the OpenJDK configurations pass). The target for jms-light/pom.xml
actually passed all tests. I do not understand the reasons for the client
tests to fail. Besides the travis configuration (see below) I have not
touched anything outside jms-light.

Changes to the travis configuration: To check compliance with the current
version of the Google Coding Style, the current versions of the checkstyle
plugins are required. One of the plugins (puppycralwer?) is now compiled
for JDK8, and older versions compiled for JDK7 are not compatible with the
current version of the Google Coding Style configuration. The coding style
is either ok or not, but this will not change between Java versions. For
that reason I propose to change the travis configuration to only run the
the unit and integration tests for the different JDKs. The compliance tests
(including the coding style test) are run once at the end with Oracle JDK8.
Note that skipping the code style checks does not work, as skipping will
still load the modules and then fail because of an
UnsupportedClassVersionError.

Cheers,
Urs

urshunkeler:
Thanks Denys! Could you also merge the pull request?

PR #58, opened 2017-03-17
Google formatting

urshunkeler:
Based on the comments from Gregory and Denys in the Project Plan (has since disappeared from the team drive?), I changed the style configuration to the official Google version and I have adapted all the formatting rules accordingly. There are no functional changes in the source code.


urshunkeler:
The pull request to test for compliance with the Google coding style fails
only on the Oracle JDK 8 configuration and for the target client/pom.xml
(the OpenJDK configurations pass). The target for jms-light/pom.xml
actually passed all tests. I do not understand the reasons for the client
tests to fail. Besides the travis configuration (see below) I have not
touched anything outside jms-light.

Changes to the travis configuration: To check compliance with the current
version of the Google Coding Style, the current versions of the checkstyle
plugins are required. One of the plugins (puppycralwer?) is now compiled
for JDK8, and older versions compiled for JDK7 are not compatible with the
current version of the Google Coding Style configuration. The coding style
is either ok or not, but this will not change between Java versions. For
that reason I propose to change the travis configuration to only run the
the unit and integration tests for the different JDKs. The compliance tests
(including the coding style test) are run once at the end with Oracle JDK8.
Note that skipping the code style checks does not work, as skipping will
still load the modules and then fail because of an
UnsupportedClassVersionError.

urshunkeler:
The configuration client/pom.xml failed in the continuous-integration/travis-ci/pr test run in Oracle JDK 7 (but not 8) and in the continuous-integration/travis-ci/push test run in Oracle JDK 8 (but not 7). In both cases the cause was a timeout:

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated

I suspect a race condition that sometimes, but not always, triggers. In fact, looking at the source code, the test which failed is known to be unstable:

(client/src/test/java/com/google/cloud/pubsub/SubscriberImplTest.java:223)
// TODO: it's really unstable test and cause of most travis failures. To solve it I decided to use sleep at the time. But we should find a real reason of it or modify requirements.

hgj:
I've spent some time into understanding what is going on. Looks like some of those tests "don't shut down properly"?

urshunkeler:
Thanks Denys! Could you also merge the pull request?

PR #52, opened 2017-03-06
PubsubProducer implementation

aechalf-g:
First iteration of implementing PubsubProducer. 
Current progress: producer is capable of sending messages that are then subscribed to using a CPS subscription. 
ProducerThreadPool can be run to try this functionality out.
PubsubConsumer is included, but nothing implemented yet.
Unit tests included only for things that did not include significant side effects.

PR #60, opened 2017-03-20
Inject mapped API into loadtest framework

aechalf-g:
This pull request is dependent on #52 and should be merged after #52 is merged. 

Purpose of PR is to inject the current implementation of the mapped API into the loadtest framework. This includes adding a new CPSPublisherTask, as defined by the steps in the wiki. 

This PR has been updated with current PubsubProducer changes, but focus for this should be on changes/additions in loadtest framework.


load-test-framework/src/main/java/com/google/pubsub/flic/output/SheetsService.java
```diff
@@ -170,6 +171,8 @@ public void sendToSheets(String sheetId, Map<ClientType, Controller.LoadtestStat
           valueRow.add(kafkaSubscriberCount);
           kafkaValues.add(valueRow);
           break;
+        case CPS_MAPPED_JAVA_PUBLISHER:
```
mdietz94: Why not create a row for this?

mdietz94:
This should not have been merged until #52 was approved, I thought.

PR #74, opened 2017-04-04
add go subscriber loadtest

pongad:


load-test-framework/src/main/java/com/google/pubsub/flic/controllers/Client.java
```diff
@@ -111,6 +111,7 @@ public static String getTopicSuffix(ClientType clientType) {
       case CPS_GCLOUD_JAVA_SUBSCRIBER:
       case CPS_GCLOUD_PYTHON_PUBLISHER:
       case CPS_GCLOUD_GO_PUBLISHER:
+      case CPS_GCLOUD_GO_SUBSCRIBER:
```
mdietz94: Since you have a publisher and subscriber you should have a new prefix, that way you can run a test comparing go e2e against Java e2e. You could use gcloud-go

pongad: Do you mean "suffix"? Done, but I'm not sure how this changes things. I could test go end-to-end before, unless you mean run both go and java end-to-end at the same time?

pongad:
@mdietz94 Saw your LGTM. If it looks good to you, please merge. I don't have write access to this repo.

Issue #77, opened 2017-04-06
stackdriver logs from pub/sub to kafka cps-source-connector


jare19:
*Kafka Version 2.11-0.10.1.0 (have also tried 2.11-0.10.2.0 with updates to the deps)*
I have hit a road block here, I have ran though the docs too many times to count and still am coming up short.  Hopefully someone can give me a pointer here so I can get this working.
## Issue
When running the cps-source-connector I get this error:
```
ERROR Task CPSConnector-3 threw an uncaught and unrecoverable exception (org.apache.kafka.connect.runtime.WorkerTask:142)
org.apache.kafka.connect.errors.DataException: Conversion error: null value for field that is required and has no default value
	at org.apache.kafka.connect.json.JsonConverter.convertToJson(JsonConverter.java:556)
	at org.apache.kafka.connect.json.JsonConverter.convertToJsonWithEnvelope(JsonConverter.java:537)
	at org.apache.kafka.connect.json.JsonConverter.fromConnectData(JsonConverter.java:291)
	at org.apache.kafka.connect.runtime.WorkerSourceTask.sendRecords(WorkerSourceTask.java:182)
	at org.apache.kafka.connect.runtime.WorkerSourceTask.execute(WorkerSourceTask.java:160)
	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:140)
	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:175)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```
## Steps to Reproduce
1. Create pub/sub topic and subcription
2. Create stackdriver export (sink) with a basic log filter, for example: mine is just `gce_instance`
3.  In GCE instance install/start kafka via quickstart guide 
4. Follow README instructions from this kafka pubsub repo to get the source connector setup and configured.
5. create kafka topic
6.  run `bin/connect-standalone.sh config/connect-standalone.properties config/cps-source-connector.properties` to start connector
*properties file looks like this*
```
name=CPSConnector
[...]

PR #78, opened 2017-04-06
Fix null key and exception about read-only ByteBuffer that happens wh…

kamalaboulhosn:
…en using a JsonConverter in a source connector.

Fixes issue #77.

### DIFF
--- a/jms-light/src/main/java/com/google/pubsub/jms/light/destination/PubSubTemporaryTopic.java
+++ b/jms-light/src/main/java/com/google/pubsub/jms/light/destination/PubSubTemporaryTopic.java
@@ -1,0 +1,33 @@
+package com.google.pubsub.jms.light.destination;
+
+import javax.jms.JMSException;
+import javax.jms.TemporaryTopic;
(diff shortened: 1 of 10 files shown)


jare19:
I tested a build with maven including this PR and it does not complete (I see one check has failed here too) 
```
../cloud-pubsub-kafka/kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java:[65,58] error: incompatible types: Set<Object> cannot be converted to Set<String>
```
thanks for the work on this!

kamalaboulhosn:
Did you switch to the kafka-connector-work branch? That branch was behind some changes on master that are required when using Java 7. I have now rebased that branch and it should build with Java 7. Master with adding this change should work regardless.

jare19:
Yep, I did move your changes to master and tried from there..., I will try a fresh go at it, I probably screwed something up. THANKS!

jare19:
environment error on my end, this PR fixes the issue #78

PR #84, opened 2017-04-12
Dependency cleanup

jcustenborder:
Hello there! Thanks for putting together this connector! I made a couple quick changes to the dependencies. The kafka dependencies should be in the provided scope since they are hosted in the connect worker process. I also moved the testing dependencies to the test scope. This should reduce the overall size of the output artifact.


jcustenborder:
I signed it!

Issue #88, opened 2017-04-21
Clean up warnings


urshunkeler:
The current code compiles with warnings. Modify the maven configuration (pom.xml) to show the warnings (-Xlint:all), then clean up the warnings.

urshunkeler:
Thanks @PhyFrancis !

PR #90, opened 2017-04-21
add Xlint:all to pom config

PhyFrancis:
#88 

run ` mvn clean verify` from the jms-light directory and the warnings will show up.

### DIFF
--- a/client-samples/pom.xml
+++ b/client-samples/pom.xml
@@ -51,6 +51,9 @@
         <configuration>
           <source>1.8</source>
           <target>1.8</target>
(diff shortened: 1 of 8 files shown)


urshunkeler:
See my pull request (#91) for your branch. Also, should we change files outside the jms-light directory? My understanding is that the other directories are an independent project not under our control.

PhyFrancis:
Thanks @urshunkeler ! I realized that we are not actively working on the other directories, so I have limited the scope of this PR to be within jms-light.

PR #106, opened 2017-06-28
Adding ${os.detected.classifier} to Pom

dsdinter:
Just adding ${os.detected.classifier} to Pom so the right netty-tcnative-boring-ssl-static library is selected.
This is to ensure the connector can work in any OS (e.g. OSX).


dsdinter:
Triggered by mistake.

PR #114, opened 2017-08-12
Mapped API Producer 

gewillovic:


pubsub-mapped-api/src/test/java/com/google/pubsub/clients/producer/ConfigTest.java
```diff
@@ -1,18 +1,24 @@
...
+import java.util.Properties;
+
+import com.google.common.collect.ImmutableMap;
+
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.serialization.Serializer;
+import org.apache.kafka.common.serialization.StringSerializer;
+
+/**
+ * Created by gewillovic on 8/1/17.
```
kamalaboulhosn: Remove these in all the files.

pubsub-mapped-api/src/test/java/com/google/pubsub/clients/producer/ConfigTest.java
```diff
@@ -1,18 +1,24 @@
 package com.google.pubsub.clients.producer;
```
kamalaboulhosn: Make sure the license text is at the top of every source file.

pubsub-mapped-api/src/test/java/com/google/pubsub/clients/producer/ProducerTest.java
```diff
@@ -0,0 +1,215 @@
...
+  @Test
+  public void testFlush() {
+    publisher.send(new ProducerRecord<String, Integer>("topic", 123));
+
+    publisher.flush();
+
+    publisher.send(new ProducerRecord<String, Integer>("topic", 456));
+
+    try {
+      Mockito.verify(stub, Mockito.times(1)).
```
kamalaboulhosn: I believe that many of these Mockito.verify... commands will fit on single lines.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+package com.google.pubsub.clients.producer;
+
+import com.google.api.core.ApiFuture;
+
```
gewillovic: removed

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+    lock = new Semaphore(1, true);
+    closed = new AtomicBoolean(false);
+    retries = configs.getInt(ProducerConfig.RETRIES_CONFIG);
+    retriesMs = configs.getLong(ProducerConfig.RETRY_BACKOFF_MS_CONFIG);
+    lingerMs = configs.getLong(ProducerConfig.LINGER_MS_CONFIG);
+    batchSize = configs.getInt(ProducerConfig.BATCH_SIZE_CONFIG);
+    autoCreate = configs.getBoolean(ProducerConfig.AUTO_CREATE_CONFIG);
+    elementCount = configs.getLong(ProducerConfig.ELEMENTS_COUNT_CONFIG);
+    isAcks = configs.getString(ProducerConfig.ACKS_CONFIG).matches("1|all");
```
gewillovic: true, sorry missed it.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+              .setDelayThreshold(Duration.ofMillis(lingerMs))
+              .build())
+          .setRetrySettings(RetrySettings.newBuilder()
+              .setMaxAttempts(retries)
+              .setRetryDelayMultiplier(MULTIPLIER)
+              .setInitialRetryDelay(Duration.ofMillis(retriesMs))
+              .setMaxRetryDelay(Duration.ofMillis((retries + 1) * retriesMs))
+              .setRpcTimeoutMultiplier(MULTIPLIER)
+              .setInitialRpcTimeout(Duration.ofSeconds(TOTAL_DELAY))
+              .setMaxRpcTimeout(Duration.ofSeconds(TOTAL_DELAY))
```
gewillovic: But, yeah not the best way to do it, I will look into it again.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+  public Future<RecordMetadata> send(ProducerRecord<K, V> record, Callback callback) {
+    if (closed.get())
+      throw new IllegalStateException("Cannot send after the producer is closed.");
+
+    if (record == null)
+      throw new NullPointerException();
+
+    final Callback cb = callback;
+
+    byte[] valueBytes = ByteString.EMPTY.toByteArray();
```
gewillovic: I needed to before adding the check, not of any use right now.
Removed.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+            " specified in value.serializer");
+      }
+    }
+
+    if (valueBytes.length == 0)
+      throw new NullPointerException("Value cannot be null or an empty string.");
+
+    Map<String, String> attributes = new HashMap<>();
+
+    byte[] keyBytes = ByteString.EMPTY.toByteArray();
```
gewillovic: Removed.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+      try {
+        keyBytes = keySerializer.serialize(record.topic(), record.key());
+      } catch (ClassCastException cce) {
+        throw new SerializationException("Can't convert key of class " +
+            record.key().getClass().getName() + " to class " +
+            producerConfig.getClass(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG).getName() +
+            " specified in key.serializer");
+      }
+    }
+    attributes.put("key", new String(keyBytes));
```
gewillovic: Yeah, I figured this out while pulling through the console - but didn't know which option to go with -.
What is the appropriate thing to do if the key exceeded this threshold? I tried to find any analogous config in Kafka, but it's all about the message size (payload).

pubsub-mapped-api/src/test/java/com/google/pubsub/clients/producer/ConfigTest.java
gewillovic: Removed.

pubsub-mapped-api/src/test/java/com/google/pubsub/clients/producer/ConfigTest.java
gewillovic: Done.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ProducerConfig.java
```diff
@@ -0,0 +1,165 @@
...
+
+  ProducerConfig(Map<?, ?> props) {
+    super(CONFIG, props);
+  }
+
+  ProducerConfig(Map<?, ?> props, boolean doLog) { super(CONFIG, props, doLog); }
+
+  public static Set<String> configNames() { return CONFIG.names(); }
+
+  public static void main(String[] args) { System.out.println(CONFIG.toHtmlTable()); }
```
gewillovic: Yeah, we may need to discuss this class in detail.

pubsub-mapped-api/src/main/java/com/google/pubsub/driver/Main.java
```diff
@@ -0,0 +1,50 @@
+package com.google.pubsub.driver;
+
+import java.util.Properties;
+
+import com.google.pubsub.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+
+/**
+ * Created by gewillovic on 8/4/17.
```
gewillovic: True, I forgot to remove it before creating the PR.
I think we will need to provide better example classes, this was meant to be a simple driver.
Removed.

pubsub-mapped-api/src/test/java/com/google/pubsub/clients/producer/ProducerTest.java
gewillovic: Fixed.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    SettableFuture<RecordMetadata> future = SettableFuture.create();
+    future.set(recordMetadata);
+    return future;
+  }
+
+  private void checkRecordSize(int size) {
+    if (size > batchSize) {
+      throw new BufferExhaustedException("Message is " + size +
```
gewillovic: I got it wrong then. Yeah, it is more plausible this way.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+            cb.onCompletion(recordMetadata, null);
+          }
+        });
+      } else {
+        callback.onCompletion(recordMetadata, null);
+      }
+    }
+
+    SettableFuture<RecordMetadata> future = SettableFuture.create();
+    future.set(recordMetadata);
```
gewillovic: Done.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: Done, check them again, if you please.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,364 @@
...
+              .build())
+          .setRetrySettings(RetrySettings.newBuilder()
+              .setMaxAttempts(retries)
+              .setRetryDelayMultiplier(MULTIPLIER)
+              .setInitialRetryDelay(Duration.ofMillis(retriesMs))
+              .setMaxRetryDelay(Duration.ofMillis((retries + 1) * retriesMs))
+              .setRpcTimeoutMultiplier(MULTIPLIER)
+              .setInitialRpcTimeout(Duration.ofSeconds(TOTAL_DELAY))
+              .setMaxRpcTimeout(Duration.ofSeconds(TOTAL_DELAY))
+              .setTotalTimeout(Duration.ofSeconds(TOTAL_DELAY))
```
gewillovic: There is a semantic difference concerning requests and batches.
I will use it for now, but there maybe a better way to manipulate it.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
kamalaboulhosn: For now, let's just treat it as an error and return it to the user.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/config/ProducerConfig.java
```diff
@@ -0,0 +1,227 @@
...
+import org.apache.kafka.common.config.ConfigDef.Type;
+import org.apache.kafka.common.config.ConfigDef.Range;
+import org.apache.kafka.common.serialization.Serializer;
+import org.apache.kafka.common.config.ConfigDef.Importance;
+import org.apache.kafka.common.config.ConfigDef.ValidString;
+
+/**
+ * Provides the configurations for a KafkaProducer instance.
+ */
+public class ProducerConfig extends AbstractConfig {
```
kamalaboulhosn: Are we ultimately only going to include the properties that are relevant for CPS? If I understand correctly, on the subscriber side, we are only adding the CPS-exclusive properties.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,398 @@
...
+    List<ProducerInterceptor<K, V>> interceptorList =
+        (List) (new ProducerConfig(configs.originals())).getConfiguredInstances(
+            ProducerConfig.INTERCEPTOR_CLASSES_CONFIG, ProducerInterceptor.class);
+    this.interceptors =
+        interceptorList.isEmpty() ? null : new ProducerInterceptors<>(interceptorList);
+    
+    publishers = new ConcurrentHashMap<>();
+  }
+
+  //TODO: deal with these magic numbers.
```
kamalaboulhosn: Which magic numbers?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,398 @@
...
+    TopicName topicName = null;
+    TopicAdminClient topicAdmin = null;
+
+    try {
+      topicName = TopicName.create(project, topic);
+      topicAdmin = TopicAdminClient.create();
+      topicAdmin.getTopic(topicName);
+    } catch (Exception e) {
+      if (autoCreate)
+        topicAdmin.createTopic(topicName);
```
kamalaboulhosn: Do you want to create the topic regardless of Exception? I'd think that only the NotFound exception would be the one on which we'd want to create the topic.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,398 @@
...
+   * The given record must have the same topic as the producer.
+   */
+  public Future<RecordMetadata> send(ProducerRecord<K, V> originalRecord, Callback callback) {
+    if (closed.get())
+      throw new IllegalStateException("Cannot send after the producer is closed.");
+
+    ProducerRecord<K, V> record =
+        this.interceptors == null ? originalRecord : this.interceptors.onSend(originalRecord);
+
+    if (record == null)
```
kamalaboulhosn: Please use braces even when the body only has a single statement.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,398 @@
...
+            record.key().getClass().getName() + " to class " +
+            producerConfig.getClass(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG).getName() +
+            " specified in key.serializer");
+      }
+    }
+    attributes.put("key", keyBytes == null ? "" : new String(Base64.encodeBase64(keyBytes)));
+
+    PubsubMessage msg = PubsubMessage.newBuilder()
+        .setData(ByteString.copyFrom(valueBytes))
+        .putAllAttributes(attributes)
```
kamalaboulhosn: Does this mean we are going to depend on the server to reject messages that end up with a key larger than 1024 bytes?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,398 @@
...
+        .build();
+
+    ApiFuture<String> messageIdFuture = null;
+
+    messageIdFuture =
+        publishers.computeIfAbsent(
+            record.topic(), K -> new AtomicReference<>(createPublisher(record.topic())))
```
kamalaboulhosn: You can rename "K" to "topic" and then just use it in place of record.topic(). Is the AtomicReference necessary?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,398 @@
...
+    if (callback != null) {
+      if (isAcks) {
+        ApiFutures.addCallback(messageIdFuture, new ApiFutureCallback<String>() {
+          private RecordMetadata recordMetadata =
+              getRecordMetadata(topic, 0L, keySize, valueSize);
+
+          @Override
+          public void onFailure(Throwable t) {
+            callbackOnCompletion(cb, recordMetadata, new ExecutionException(t));
+            future.set(recordMetadata);
```
kamalaboulhosn: In this case, I think you want to call setException on the future.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/config/ProducerConfig.java
gewillovic: I am working on it right now, just waiting for the subscriber's PR to get approved.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: The " + 1 ", " + 2 " thing, nevermind I think they are good.
It's a TODO for me to check later.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: Okey, will do it.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/config/ProducerConfig.java
gewillovic: I created a separate branch with the refactored configs, utilizing Kafka's configs as is, and CPS's as a separate layer over it.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: True.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: Done.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: Nope, fixed.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: True, fixed.

load-test-framework/src/main/java/com/google/pubsub/clients/mapped/CPSPublisherTask.java
```diff
@@ -25,7 +24,7 @@
   private final String topic;
   private final String payload;
   private final int batchSize;
-  private final PubsubProducer<String, String> publisher;
+  private final KafkaProducer<Double, String> publisher;
```
kamalaboulhosn: Why is this now a Double?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.config.ConfigDef.Type;
+import org.apache.kafka.common.config.ConfigDef.Range;
+import org.apache.kafka.common.config.ConfigDef.Importance;
+
+/**
+ * Provides the configurations for a Kafka Producer instance.
+ */
+@VisibleForTesting
+public class ExtendedConfig {
```
kamalaboulhosn: Let's be consistent in naming and layout with the consumer. That means renaming PubSubConfig to PubSubProducerConfig and ExtendedConfig to Config.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public static final String PROJECT_CONFIG = "project";
+    private static final String PROJECT_DOC = "GCP project that we will connect to.";
+
+    public static final String ELEMENTS_COUNT_CONFIG = "element.count";
+    private static final String ELEMENTS_COUNT_DOC = "This configuration controls the default count of"
+            + " elements in a batch.";
+
+    public static final String AUTO_CREATE_CONFIG = "auto.create.topics.enable";
+    private static final String AUTO_CREATE_DOC = "A flag, when true topics are automatically created"
```
kamalaboulhosn: No need to say "A flag"

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
```diff
@@ -0,0 +1,81 @@
...
+  public static class PubSubConfig extends AbstractConfig {
+
+    public static final String PROJECT_CONFIG = "project";
+    private static final String PROJECT_DOC = "GCP project that we will connect to.";
+
+    public static final String ELEMENTS_COUNT_CONFIG = "element.count";
+    private static final String ELEMENTS_COUNT_DOC = "This configuration controls the default count of"
+            + " elements in a batch.";
+
+    public static final String AUTO_CREATE_CONFIG = "auto.create.topics.enable";
```
kamalaboulhosn: Let's be consistent with the Consumer here and call this topic.allow.create.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,424 @@
...
+
+    attributes.put("id", clientId);
+
+    attributes.put("timestamp", dateTime);
+
+    attributes.put("version", Long.toString(VERSION));
+
+    attributes.put("key", key == null ? "" : new String(Base64.encodeBase64(key)));
+
+    if (attributes.get("key").getBytes().length > 1024) {
```
kamalaboulhosn: Don't do the getBytes(). This introduces platform-specific encoding. Since we are already base64 encoded, should be able to just get the length of the string.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
```diff
@@ -0,0 +1,424 @@
...
+    final String topic = record.topic();
+    final int keySize = keyBytes == null ? 0 : keyBytes.length, valueSize = valueBytes.length;
+
+    final SettableFuture<RecordMetadata> future = SettableFuture.create();
+
+    if (callback != null) {
+      if (isAcks) {
+        ApiFutures.addCallback(messageIdFuture, new ApiFutureCallback<String>() {
+          private RecordMetadata recordMetadata =
+              getRecordMetadata(topic, dateTime.getMillis(), keySize, valueSize);
```
kamalaboulhosn: Shouldn't the offset passed into here match the timestamp that was set for the "timestamp" attribute of the message? Probably worth checking in a unit test.

load-test-framework/src/main/java/com/google/pubsub/clients/mapped/CPSPublisherTask.java
gewillovic: It's not of any importance, I will change the task completely anyway.
I just copied the initializations from another class.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
gewillovic: Okay, but I think for the pubsub configs to be an inner class is the proper way - as it serves for nothing on its own -, isn't it?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
gewillovic: I copied it from the brokers' configs, should I change it?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
gewillovic: Done.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: Isn't it set by the server side? please elaborate.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
kamalaboulhosn: I leave it to you and @pietrzykp to decide how you want to do it. If we aren't going to access the config in any way, then an inner class may be the best. But let's just make sure we are consistent across the Producer and Consumer.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
kamalaboulhosn: @pietrzykp, can we update the consumer code to use the same format in that case?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
kamalaboulhosn: Actually, I was mistaken. You do use the same dateTime object for the recordMetadata and the timestamp attribute. However, I'd recommend passing the dateTime attribute as something other than a string to createMessage. I think using the milliseconds for it makes the most sense. And in fact, let's rename "timestamp" to "offset." The fact that we are using the timestamp is not necessarily the interesting part. It is that we are providing an offset that is consistent across the publisher and subscriber.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
pietrzykp: Okay, will do - but for me this is about creating subscriptions. "auto.create.subscriptions.enable" then?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/ExtendedConfig.java
pietrzykp: Inner class makes a lot of sense. Though, I feel like when user will be scanning through our files it would be easier for them to understand if it looks just as Kafka's configuration file they've already seen (like very-analogical, separate file). I'm fine with both ways.

pietrzykp: Also, my configuration is built entirely differently - in constructor it populates its attributes, so all of "consumerConfig.getString(ConsumerConfig.GROUP_ID_CONFIG)" are done there. It allows me not to keep this huge list of config attributes in Consumer implementation, rather then that I keep "config" object and just call config.getGroupId(). It also makes easier to distinguish between attributes that are configuration and different attributes of consumer.

gewillovic: I think the user is not concerned about "our" files, he should be more concerned about any changes he will need to make - to his code -, which is the same in both cases. At the end of the day, they will just look at our documentation - which still, won't differ.

gewillovic: Yeah, I have checked your configuration, it's elegant but I targeted a simple transition for now. It's on my TODO list.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: Done.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/producer/KafkaProducer.java
gewillovic: Done.

PR #115, opened 2017-08-14
Resolve dependency issue

jinankjain:
Resolve this issue

```
Exception in thread "main" java.lang.NoSuchMethodError: com.google.common.collect.Sets$SetView.iterator()Lcom/google/common/collect/UnmodifiableIterator;
	at org.reflections.Reflections.expandSuperTypes(Reflections.java:380)
	at org.reflections.Reflections.<init>(Reflections.java:126)
	at org.apache.kafka.connect.runtime.isolation.DelegatingClassLoader.scanPluginPath(DelegatingClassLoader.java:221)
	at org.apache.kafka.connect.runtime.isolation.DelegatingClassLoader.scanUrlsAndAddPlugins(DelegatingClassLoader.java:198)
	at org.apache.kafka.connect.runtime.isolation.DelegatingClassLoader.initLoaders(DelegatingClassLoader.java:159)
	at org.apache.kafka.connect.runtime.isolation.Plugins.<init>(Plugins.java:47)
	at org.apache.kafka.connect.cli.ConnectStandalone.main(ConnectStandalone.java:68)
```

### DIFF
--- a/kafka-connector/pom.xml
+++ b/kafka-connector/pom.xml
@@ -13,6 +13,11 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+       <version>23.0</version>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>


jinankjain:
I signed it!

kamalaboulhosn:
I'm a little surprised this is necessary. I don't think the connector refers to anything in Guava that isn't already pulled in with the dependency on com.google.apis (which is used for ListenableFuture as it is part of the API used to make Pull calls). Do you have steps to reproduce this issue?

jinankjain:
@kamalaboulhosn 

Steps to reproduce:

I ran these commands ./bin/connect-distributed.sh config/connect-distributed.properties /kafka-connector/config/cps-sink-connector.properties. CLASSPATH was set properly to the jar generated by mvn package.

kamalaboulhosn:
@jinankjain Which version of Kafka were you running and on which platform?

jinankjain:
@kamalaboulhosn `kafka_2.12-0.11.0.0`

kamalaboulhosn:
Thank you for pointing this issue out. I think PR #117 is a little bit better approach. Ultimately, the issue is that some of the connector's dependencies pull in an older version of guava. Rather than add an explicit dependency to the connector that it ultimately doesn't use, I think it's better just to exclude guava from the built package.

jinankjain:
Agreed 👍

PR #117, opened 2017-08-16
Shade the guava library to avoid a NoMethodFoundError 

kamalaboulhosn:
A slightly better approach than PR #115, which adds the (unneeded) guava dependency.


jinankjain:
Awesome thanks for the patch and help 👍

PR #119, opened 2017-08-22
Fix some broken markdown in the README

jamiew:
Some newlines snuck their way into these links

Thanks for the project!


jamiew:
I just signed the CLA

PR #122, opened 2017-09-15
Mapped api ack

pietrzykp:
Commit functionality, refreshing ack deadlines and acknowledging.


pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/KafkaConsumer.java
```diff
@@ -348,21 +356,24 @@ private void checkPatternSubscribePreconditions(Pattern pattern) {
...
+    for(Subscriber s: topicNameToSubscriber.values()) {
+      s.stopAsync().awaitTerminated();
+    }
+    topicNameToSubscriber = ImmutableMap.of();
     topicNames = ImmutableList.of();
     currentPoolIndex = 0;
   }
 
-  private void deleteSubscriptionsIfAllowed(Collection<Subscription> subscriptions) {
+  private void deleteSubscriptionsIfAllowed(Collection<Subscriber> subscribers) {
```
mdietz94: Why subscription to subscriber?

pietrzykp: True, changed back.

PR #124, opened 2017-09-27
increase parallel puller count

pongad:
The new default number of parallel pullers is NumCPU.
To keep up with publishers, we need to increase the number of pullers.
This commit increaes the number to 5*NumCPU,
the same as the old default.

### DIFF
--- a/load-test-framework/src/main/java/com/google/pubsub/clients/gcloud/CPSSubscriberTask.java
+++ b/load-test-framework/src/main/java/com/google/pubsub/clients/gcloud/CPSSubscriberTask.java
@@ -39,10 +39,13 @@ class CPSSubscriberTask extends Task implements MessageReceiver {
 
   private CPSSubscriberTask(StartRequest request) {
     super(request, "gcloud", MetricsHandler.MetricName.END_TO_END_LATENCY);
-    this.subscription = SubscriptionName.create(
-        request.getProject(), request.getPubsubOptions().getSubscription());
+    this.subscription =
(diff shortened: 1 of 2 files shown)


load-test-framework/src/main/java/com/google/pubsub/clients/gcloud/CPSSubscriberTask.java
```diff
@@ -39,10 +39,13 @@
...
     super(request, "gcloud", MetricsHandler.MetricName.END_TO_END_LATENCY);
-    this.subscription = SubscriptionName.create(
-        request.getProject(), request.getPubsubOptions().getSubscription());
+    this.subscription =
+        SubscriptionName.create(request.getProject(), request.getPubsubOptions().getSubscription());
     try {
-      this.subscriber = Subscriber.defaultBuilder(this.subscription, this).build();
+      this.subscriber =
+          Subscriber.defaultBuilder(this.subscription, this)
+              .setParallelPullCount(Runtime.getRuntime().availableProcessors() * 5)
```
mdietz94: Why is this called parallel pull count when it is doing StreamingPull?

pongad: I'm not quite sure what you're asking. We are using StreamingPull and the number determines how many connections we create to pull in parallel.

pongad:
If this looks good to you, could you merge? I don't have write access :)

pongad:
ping

PR #125, opened 2017-09-27
java: don't sync on shutdown

pongad:
Shutting down needs to wait for messages to be processed,
and processing the messages need to lock the Task object.
So, if the thread calling shutdown locks the Task object, we'll
deadlock.


mdietz94:
I think we should add a boolean for shuttingDown, and then return early in doRun if that does happen, since otherwise we could potentially have doRun called after shutdown, since the doRuns are scheduled, which would then end up reinstantiating the subscriber.

pongad:
@mdietz94 If this looks good to you please merge. I don't have write access :)

mdietz94:
Done, thanks

PR #133, opened 2017-10-13
Fix consumer throttle.

gewillovic:
Fixed the performance hit caused by auto/manual commit feature on consumer side.

### DIFF
--- a/pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/KafkaConsumer.java
+++ b/pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/KafkaConsumer.java
@@ -113,9 +113,9 @@ public class KafkaConsumer<K, V> implements Consumer<K, V> {
   private static final ConsumerRebalanceListener NO_REBALANCE_LISTENER = null;
   private static final Logger log = LoggerFactory.getLogger(KafkaConsumer.class);
 
+  private final String GOOGLE_CLOUD_PROJECT;
   private final String TOPIC_PREFIX;
   private final String SUBSCRIPTION_PREFIX;
-  private final String GOOGLE_CLOUD_PROJECT;
 
   //because of no partition concept, the partition number is constant
   private static final int DEFAULT_PARTITION = 0;
@@ -182,14 +182,15 @@ public class KafkaConsumer<K, V> implements Consumer<K, V> {
 
       Preconditions.checkNotNull(channel);
 
+      // 32 MB for inbound and outbound messages
       SubscriberFutureStub subscriberFutureStub = SubscriberGrpc.newFutureStub(channel)
-          .withMaxInboundMessageSize(16777216)
-          .withMaxOutboundMessageSize(16777216)
+          .withMaxInboundMessageSize(32 * 1024 * 1024)
+          .withMaxOutboundMessageSize(32 * 1024 * 1024)
           .withDeadlineAfter(config.getRequestTimeoutMs(), TimeUnit.MILLISECONDS);
 
       PublisherFutureStub publisherFutureStub = PublisherGrpc.newFutureStub(channel)
-          .withMaxInboundMessageSize(16777216)
-          .withMaxOutboundMessageSize(16777216)
+          .withMaxInboundMessageSize(32 * 1024 * 1024)
+          .withMaxOutboundMessageSize(32 * 1024 * 1024)
           .withDeadlineAfter(config.getRequestTimeoutMs(), TimeUnit.MILLISECONDS);
 
       if (callCredentials != null) {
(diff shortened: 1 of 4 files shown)


pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/ack/MessageDispatcher.java
```diff
@@ -76,9 +75,9 @@
   private final FlowController flowController;
   private final MessageWaiter messagesWaiter;
 
+  private Set<String> pendingNacks;
```
pietrzykp: it would be nice to say here explicitly that these are concurrent (thread safe).

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/ack/MessageDispatcher.java
```diff
@@ -542,22 +535,20 @@ private void setupNextAckDeadlineExtensionAlarm(Instant expiration) {
...
-        }
-        logger.log(Level.FINER, "Sending {0} acks", acksToSend.size());
-      }
+      acksToSend = pendingAcks;
+      pendingAcks = new ConcurrentHashMap<>();
     }
+    logger.log(Level.FINER, "Sending {0} acks", acksToSend.size());
 
-    if(offset != null) {
+    if (offset != null) {
```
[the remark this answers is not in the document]
pietrzykp: this will not work. after you call filterAcksBeforeOffset all messages that are not before this offset will never be acked and their deadlines will be still extended. in this case you would have to put all removed values back to pendingAcks (as it is empty now).

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/ack/MessageDispatcher.java
```diff
@@ -631,20 +624,22 @@ private void extendAckDeadlines(List<PendingModifyAckDeadline> ackDeadlineExtens
...
-          for (String ackId : pendingNacks) {
-            nacksToSend.addAckId(ackId);
-          }
-          logger.log(Level.FINER, "Sending {0} nacks", pendingNacks.size());
-        } finally {
-          pendingNacks.clear();
-        }
-        modifyAckDeadlinesToSend.add(nacksToSend);
+      acksToExtend = pendingNacks;
+      pendingNacks = ConcurrentHashMap.newKeySet();
```
pietrzykp: what if sendAckOperations throws an error? are acksToSend put back in pendingNacks, or are they lost?

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/ack/MessageDispatcher.java
gewillovic: Okay.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/ack/MessageDispatcher.java
```diff
@@ -542,22 +535,20 @@ private void setupNextAckDeadlineExtensionAlarm(Instant expiration) {
     }
   }
 
+  //TODO: do we really need the sync?
   void acknowledgePendingMessages(boolean sync, Long offset) {
-    Map<String, AckHandler> acksToSend = new HashMap<>();
+    Map<String, AckHandler> acksToSend;
     synchronized (pendingAcks) {
```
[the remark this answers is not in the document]
gewillovic: Not sure, I am not even sure if it's of any use.
@kamalaboulhosn I will remove it, but please take a look.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/ack/MessageDispatcher.java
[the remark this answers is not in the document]
gewillovic: Fixed, check it again now.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/ack/MessageDispatcher.java
gewillovic: I don't understand what was happening before, I mean pendingNacks was being cleared anyway, right? If it's the case, then I am doing the same. I think I didn't change the logic here - or I understand the try - finally wrong then.

pubsub-mapped-api/src/main/java/com/google/pubsub/clients/consumer/KafkaConsumer.java
```diff
@@ -182,14 +182,15 @@ private KafkaConsumer(Config configOptions) {
 
       Preconditions.checkNotNull(channel);
 
+      // 16 MB for inbound and outbound messages
```
mdietz94: Should be 20 MB

PR #135, opened 2017-10-18
No duplicates in Kafka when ack fails

niksajakovljevic:
The idea is not to produce duplicates in Kafka if Pub/sub ack fails. We can achieve that by managing ack ids in callbacks.

### DIFF
--- a/kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
+++ b/kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
@@ -64,6 +64,7 @@ public class CloudPubSubSourceTask extends SourceTask {
   // Keep track of all ack ids that have not been sent correctly acked yet.
   private Set<String> ackIds = Collections.synchronizedSet(new HashSet<String>());
   private CloudPubSubSubscriber subscriber;
+  private Set<String> ackIdsInFlight = Collections.synchronizedSet(new HashSet<String>());
 
   public CloudPubSubSourceTask() {}
 
@@ -120,8 +121,8 @@ public class CloudPubSubSourceTask extends SourceTask {
         PubsubMessage message = rm.getMessage();
         String ackId = rm.getAckId();
         // If we are receiving this message a second (or more) times because the ack for it failed
-        // then do not create a SourceRecord for this message.
-        if (ackIds.contains(ackId)) {
+        // then do not create a SourceRecord for this message. In case we are waiting for ack response we also skip the message
+        if (ackIds.contains(ackId) || ackIdsInFlight.contains(ackId)) {
           continue;
         }
         ackIds.add(ackId);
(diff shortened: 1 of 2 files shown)


niksajakovljevic:
I signed it!

niksajakovljevic:
I signed it again!

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
```diff
@@ -197,21 +197,28 @@ private void ackMessages() {
     if (ackIds.size() != 0) {
       AcknowledgeRequest.Builder requestBuilder = AcknowledgeRequest.newBuilder()
           .setSubscription(cpsSubscription);
+      final Set<String> ackIdsCopy = new HashSet<>();
       synchronized (ackIds) {
         requestBuilder.addAllAckIds(ackIds);
-        ackIds.clear();
+        ackIdsCopy.addAll(ackIds);
```
kamalaboulhosn: It seems like this could result in repeatedly sending acks for the same messages. While a call to ackMessages is outstanding, if another poll occurs, then all acks will be sent again. Perhaps it would be better to keep a "acksInFlight" HashSet that we move the items from ackIds to. Then, in poll, we exclude messages that are in ackIds or in acksInFlight. Once the Future returns here, we remove the items from acksInFlight.

niksajakovljevic: Thanks for the feedback. I was aware of the trade off ( sending repeated acks ) but by looking at the `SubscriberGrpc` java docs, repeated acks should not result in errors on Google side. Anyways, having global variable would eliminate repeated acks so lets go for that

niksajakovljevic:
@kamalaboulhosn I added recommended changes to prevent repeated ack ids being sent to Google servers

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
```diff
@@ -120,8 +121,8 @@ public void start(Map<String, String> props) {
         PubsubMessage message = rm.getMessage();
         String ackId = rm.getAckId();
         // If we are receiving this message a second (or more) times because the ack for it failed
-        // then do not create a SourceRecord for this message.
-        if (ackIds.contains(ackId)) {
+        // then do not create a SourceRecord for this message. In case we are waiting for ack response we also skip the message
```
kamalaboulhosn: nit: Wrap this at 100-character line width.

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
```diff
@@ -190,15 +191,17 @@ public void start(Map<String, String> props) {
   }
 
   /**
-   * Attempt to ack all ids in {@link #ackIds}. Acks are best-effort, so if acking fails, messages
```
kamalaboulhosn: I wouldn't remove this comment. Acks still are best effort since Cloud Pub/Sub has an at-least-once delivery guarantee, so duplicates are still possible.

kamalaboulhosn:
Thanks for making these changes!

PR #136, opened 2017-10-26
Acknowledge Cloud Pub/Sub messages in commit hook

niksajakovljevic:
Current implementation acks pub/sub messages in the beginning of pull method which means that it takes two pull cycles to send messages to Kafka and acknowledge them after. Whenever you stop Kafka connector, the messages from the last cycle would get delivered to Kafka but acknowledge of pub/sub messages would not happen. This produces ( cps.maxBatchSize * tasks.max) duplicates on every stop of Kafka connector. 
Provided implementation uses SourceTask commit hook to acknowledge pub/sub messages.


kamalaboulhosn:
Wonderful, thanks!

PR #139, opened 2017-11-02
Configurable converter to support different serialization formats

niksajakovljevic:
Current implementation of `CloudPubSubSinkConnector` has limited support for serialization.The idea is to enable different serialization formats (eg. Json, Avro) and to decouple serialization from Sink connector. There are already available converters for different formats that can be easily plugged in and avoid implementing serialization logic within sink connector.

### DIFF
--- a/kafka-connector/config/cps-sink-connector.properties
+++ b/kafka-connector/config/cps-sink-connector.properties
@@ -4,3 +4,5 @@ tasks.max=10
 topics=my-kafka-topic
 cps.topic=my-cps-topic
 cps.project=my-cps-project
+value.converter=com.google.pubsub.kafka.sink.CloudPubSubSinkConnectorTest$SillyStringConverter
+value.converter.prefix=somePrefix_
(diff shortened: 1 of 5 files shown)


niksajakovljevic:
My general idea was to enable any kind of serialization format you might be needing and that can mean serializing to the same format or to the new format. I totally agree that if format is staying the same then the best is to go with byte array, but in a case where you need to do some transformation on the records then it would force you to deserialize/serialize records in transformer (`org.apache.kafka.connect.transforms.Transformation`). 
You made a good point about not enforcing Kafka Connect dependency for someone who is not using Kafka. I guess I was focusing on my use case only :) 

I still think that would be nice to decouple serialization from Sink connector (and being able to supply implementation). Since Kafka Connect is going through all the records any way, to me it feels natural to perform some transformations if needed.  However I am not strongly opinionated on this one. As I would need to focus on some other things we can maybe close this pull request and open a new one if needed ?

kamalaboulhosn:
Yeah, I think we need to look at serialization more generally and how it fits in the context of the Pub/Sub connector. If we add it to the sink connector, we probably need to add it to the source connector as well. You can feel free to put in an issue to track improved serialization and we can discuss what we want to accomplish and how we might approach it. Thanks for your continued improvements to the connector!

PR #144, opened 2017-12-27
Adding timestamp message attribute 

niksajakovljevic:
When moving messages from Pubsub to Kafka, Kafka message ingress time is set to Pubsub message publish time. In some cases would be more appropriate to set Kafka ingress time to some other time (eg. event time contained in the payload). This way the semantics and reasoning about time is the same across different systems ( Kafka supports traversing through message offset by time so being able to control ingress time is important for many scenarios like backup/restore )

### DIFF
--- a/kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
+++ b/kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
@@ -171,7 +171,8 @@ public class CloudPubSubSourceConnector extends SourceConnector {
             Type.STRING,
             null,
             Importance.MEDIUM,
-            "The optional Cloud Pub/Sub message attribute to use as a timestamp for messages published to Kafka.")
+            "The optional Cloud Pub/Sub message attribute to use as a timestamp for messages published to Kafka."
+                + "The timestamp is Long value.")
         .define(
             KAFKA_PARTITIONS_CONFIG,
             Type.INT,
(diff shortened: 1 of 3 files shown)


niksajakovljevic:
Seems that there is a problem with Oracle JDK7 environment vars (JAVA_HOME not being set). OpenJDK7 and OracleJDK8 builds work fine.

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
```diff
@@ -165,6 +166,12 @@ public ConfigDef config() {
             null,
             Importance.MEDIUM,
             "The Cloud Pub/Sub message attribute to use as a key for messages published to Kafka.")
+        .define(
+            KAFKA_MESSAGE_TIMESTAMP_CONFIG,
+            Type.STRING,
+            null,
+            Importance.MEDIUM,
+            "The optional Cloud Pub/Sub message attribute to use as a timestamp for messages published to Kafka.")
```
kamalaboulhosn: Can you add to the description the format of the timestamp?

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
```diff
@@ -246,6 +255,15 @@ private int selectPartition(Object key, Object value) {
     }
   }
 
+  private Long getLongValue(String timestamp){
+    try {
```
kamalaboulhosn: Add

if(timestamp == null) {
  return null;
}

That way we can avoid having to throw and catch the exception for what I imagine will be the most typical use case.

kamalaboulhosn:
I am going to fix the build env so we don't have this problem and we can merge this request.

PR #147, opened 2018-01-09
Rethrow the exception from Grpc

lsenta:


kamalaboulhosn:
I believe I addressed this in #152 or possibly another commit.

Issue #151, opened 2018-03-08
Struct message body does not support Map or Struct types.


justinnaldzin:
Using both JSON and AVRO I receive this `DataException`

**Struct message body does not support Map or Struct types.**

But as I understand, Struct and Map types should be converted to String.

https://github.com/GoogleCloudPlatform/pubsub/blob/d9d1765ac17f7d86f7b71ee20ca5320832b8b800/kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java#L218

Here is an example of the JSON:
```
[...]

PR #163, opened 2018-06-06
Fix handling of optional fields in a struct in the sink connector

kamalaboulhosn:
More descriptive errors for #151


kafka-connector/src/test/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTaskTest.java
```diff
@@ -173,12 +173,52 @@ public void testStructSchema() {
...
+        .field(FIELD_STRING2, SchemaBuilder.string().optional()).build();
+    Struct val = new Struct(schema);
+    val.put(FIELD_STRING1, "tide");
+    SinkRecord record = new SinkRecord(null, -1, null, null, schema, val, -1);
+    List<SinkRecord> list = new ArrayList<>();
     list.add(record);
-    try {
-      task.put(list);
-    } catch (DataException e) { } // Expected, pass.
+    task.put(list);
```
jrheizelman: Could you add a case where the optional field is filled? I think it could be in this test method, just cover both cases with optional types.

kamalaboulhosn: Done.

Issue #171, opened 2018-07-19
Source kafka connector: reuse kafka partitioning


schizhov:
Currently there are 3 partitioning schemes provided:
- `round_robin`
- `hash_key`
- `hash_value`

First 2 are supported by kafka producer out of the box. To be more specific: if no partition explicitly specified in the constructor of a `SourceRecord` kafka partitioner will be used. Default one (`DefaultPartioner`) does round robin if no message key is provided and does murmur hash if there is a key.  This has 2 advantages over the current way of doing things:
- hashing that is consistent across vms
- no need to configure number of partitions explicitly

`hash_value` -  it is suggested to replace current vm-dependent hash based partitioning with a stable one. If we specify kafka message key as a stable hash of the message body then we can rely on kafka partitioner to do the rest and avoid the need to explicitly specify number of topic partitions.

PR #172, opened 2018-07-19
Fix partitioning

schizhov:
Fix for https://github.com/GoogleCloudPlatform/pubsub/issues/171:

- rely on kafka producer partitioner (no need to configure number of partitions)
- switch to murmur hashing which is consistent across vms

### DIFF
--- a/kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
+++ b/kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
@@ -43,7 +43,6 @@ public class CloudPubSubSourceConnector extends SourceConnector {
 
   private static final Logger log = LoggerFactory.getLogger(CloudPubSubSourceConnector.class);
 
-  public static final String KAFKA_PARTITIONS_CONFIG = "kafka.partition.count";
   public static final String KAFKA_PARTITION_SCHEME_CONFIG = "kafka.partition.scheme";
   public static final String KAFKA_MESSAGE_KEY_CONFIG = "kafka.key.attribute";
   public static final String KAFKA_MESSAGE_TIMESTAMP_CONFIG = "kafka.timestamp.attribute";
@@ -51,13 +50,11 @@ public class CloudPubSubSourceConnector extends SourceConnector {
   public static final String CPS_SUBSCRIPTION_CONFIG = "cps.subscription";
   public static final String CPS_MAX_BATCH_SIZE_CONFIG = "cps.maxBatchSize";
   public static final int DEFAULT_CPS_MAX_BATCH_SIZE = 100;
-  public static final int DEFAULT_KAFKA_PARTITIONS = 1;
-  public static final String DEFAULT_KAFKA_PARTITION_SCHEME = "round_robin";
+  public static final String DEFAULT_KAFKA_PARTITION_SCHEME = PartitionScheme.KAFKA_PRODUCER.value;
 
   /** Defines the accepted values for the {@link #KAFKA_PARTITION_SCHEME_CONFIG}. */
   public enum PartitionScheme {
-    ROUND_ROBIN("round_robin"),
-    HASH_KEY("hash_key"),
+    KAFKA_PRODUCER("kafka_producer"),
     HASH_VALUE("hash_value");
 
     private String value;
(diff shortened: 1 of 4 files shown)


schizhov:
I signed it!

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
```diff
@@ -71,10 +68,8 @@ public String toString() {
     }
 
     public static PartitionScheme getEnum(String value) {
-      if (value.equals("round_robin")) {
```
kamalaboulhosn: I don't think it's safe to just get rid of this mode. Other may already be depending on it. We can add new modes, but we shouldn't eliminate old ones, at least as long as we don't have the released code versioned.

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
```diff
@@ -59,18 +61,16 @@
...
-  // Keeps track of the current partition to publish to if the partition scheme is round robin.
-  private int currentRoundRobinPartition = -1;
   // Keep track of all ack ids that have not been sent correctly acked yet.
   private Set<String> ackIds = Collections.synchronizedSet(new HashSet<String>());
   private CloudPubSubSubscriber subscriber;
   private Set<String> ackIdsInFlight = Collections.synchronizedSet(new HashSet<String>());
   private final Set<String> standardAttributes = new HashSet<>();
 
-  public CloudPubSubSourceTask() {}
+  public CloudPubSubSourceTask() {
```
kamalaboulhosn: Can you please undo all of these formatting changes? I'd rather they not be included with a functional change so it can be clear what is relevant to the change and what isn't.

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
```diff
@@ -136,14 +134,19 @@ public void start(Map<String, String> props) {
...
-        if (timestamp == null){
+        if (timestamp == null) {
           timestamp = Timestamps.toMillis(message.getPublishTime());
         }
         ByteString messageData = message.getData();
         byte[] messageBytes = messageData.toByteArray();
 
+        String key = messageAttributes.get(kafkaMessageKeyAttribute);
+        if (key == null && kafkaPartitionScheme.equals(PartitionScheme.HASH_VALUE)) {
+          key = Integer.toString(Utils.murmur2(messageBytes));
```
kamalaboulhosn: This is a backwards-incompatible change and we should probably be careful with such a thing. If one is using the connector in any kind of distributed fashion, moving to a new version could result in inconsistent hashing. Can you make this a new partition scheme to maintain backwards compatibility?

schizhov: Well previous version is sort of incompatible with itself since `Object#hashCode` is not the same for the same object in different vms. I can revert it back though if you think fixing it is risky.

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
schizhov: Anyone who uses `round_robin` now should be able to get the same behaviour with  `kafka_producer` scheme. (Assuming they don't pass message keys and use `round_robin` at the same time)

kamalaboulhosn: But what if they have round_robin specified in their config? Do they have to switch to kafka_producer? If so, I'd prefer round_robin stay as a mode for now so we get backwards compatibility with configs.

schizhov: So I'll keep all current partition schemes with no changes and add a new `kafka_producer` scheme. Also the `round_robin` will remain default for the same reasons. BTW: I think I have to recreate the PR since I messed up with commit emails a bit.

schizhov:
Closing this PR to resolve issues with commiter email. Will open new PR shortly with the comments addressed.

PR #173, opened 2018-07-20
add kafka_partitioner partition scheme

schizhov:
New scheme relies on kafka producer to do the partitioning, which by default provides stable hashing based on message key or round robin if no key is specified. Number of partitions is automatically inferred by kafka producer.

Partially adresses https://github.com/GoogleCloudPlatform/pubsub/issues/171


schizhov:
Is there any specific reason this PR is on hold?

PR #188, opened 2018-12-20
Convert Structs to JSON When Publishing

evcallia:
- Add "cps.struct.to.json" option allowing users to publish Structs as JSON
- Update org.apache.kafka connect-api version so we can use the connect-json kafka converter

### DIFF
--- a/kafka-connector/README.md
+++ b/kafka-connector/README.md
@@ -98,6 +98,7 @@ Connector supports the following configs:
 |---------------|-------------|-----------------------|------------------------------------------------------------------------------------------------------------------------------------|
 | cps.topic | String | REQUIRED (No default) | The topic in Cloud Pub/Sub to publish to, e.g. "foo" for topic "/projects/bar/topics/foo". |
 | cps.project | String | REQUIRED (No default) | The project in Cloud Pub/Sub containing the topic, e.g. "bar" from above. |
+| cps.struct.to.json | Boolean | false | Whether to convert Structs to JSON when publishing to Pub/Sub | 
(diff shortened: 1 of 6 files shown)


evcallia:
I signed it!

kafka-connector/README.md
```diff
@@ -98,6 +98,7 @@ Connector supports the following configs:
 |---------------|-------------|-----------------------|------------------------------------------------------------------------------------------------------------------------------------|
 | cps.topic | String | REQUIRED (No default) | The topic in Cloud Pub/Sub to publish to, e.g. "foo" for topic "/projects/bar/topics/foo". |
 | cps.project | String | REQUIRED (No default) | The project in Cloud Pub/Sub containing the topic, e.g. "bar" from above. |
+| cps.struct.to.json | Boolean | false | Whether to convert Structs to JSON when publishing to Pub/Sub |
```
kamalaboulhosn: If we are going to add this, it would be better for it to be more generic. Instead of specifying JSON or not JSON, it would probably be better to allow one to specify the converter to use much like one can do in Kafka Connect itself (see value.converter in https://kafka.apache.org/documentation/#connectconfigs).

Additionally, this change might not be necessary. What is the type of the message that is coming into the connector? If it is already JSON, then one could use a StringConverter to just pass the data through. If it is some other format, then this might be necessary. This conversion inside this connector should only be necessary if the format of the message stored in Kafka differs from the format desired in the message written to Cloud Pub/Sub.

evcallia: In my case I need to convert avro to json. I see what you mean about making it more generic though.

kamalaboulhosn:
Hasn't been follow-up on on this PR, so closing for now.

PR #193, opened 2019-02-21
Fix too many acks

schizhov:
This PR fixes 2 problems:
1. When number of ack ids gets larger then fit into single AcknowledgeRequest, connector cannot ack any message, gets those back and keeps writing duplicates to kafka.
2. commit method is supposed to wait until all messages are acked. Previously it was not, which could result in extra duplicates during connector restarts.


schizhov:
Closing PR to resubmit after fixing the user

PR #195, opened 2019-04-01
Allow Kafka Connect headers

gh-mlfowler:
Version 1.1.0 introduced Headers to Kafka Connect which is used by some connectors such as Debezium. This change exposes a configuration item `headers.publish` modelled after `metadata.publish`.

PR #198, opened 2019-04-10
Update kafka-connect api to 2.1.0 and add support for ConnectHeaders

cyanezp:
Updating to kafka-api 2.1.0 allows us to use ConnectHeaders and this way match Kafka Headers to PubSub message attributes and vice versa

### DIFF
--- a/kafka-connector/pom.xml
+++ b/kafka-connector/pom.xml
@@ -59,7 +59,7 @@
     <dependency>
       <groupId>org.apache.kafka</groupId>
       <artifactId>connect-api</artifactId>
-      <version>0.10.2.1</version>
+      <version>2.1.0</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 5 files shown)


cyanezp:
I signed it!

kamalaboulhosn:
Does this do something different from #195?

cyanezp:
Oh... I'm so sorry!
I committed to my repo on March 11 but I totally forgot to create a pull request (newbie mistake). I'll diff the code to see if it's something different, but I think we're looking for the same goal.

cyanezp:
Indeed, there is a difference with #195 because in that PR only supports to copy headers from Kafka to PubSub (CloudPubSubSinkTask), but this PR also allows to copy attributes from PubSub to Kafka (CloudPubSubSourceTask). So, I'll close this PR, update my branch with the latest changes and create a new pull request.

Thanks!

PR #200, opened 2019-04-10
Copy Kafka headers to PubSub attributes

cyanezp:
Same as previous failed attempt, I'm generating this pull request to allow to copy Kafka attributes to PubSub headers (like PR #195)


kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -159,6 +162,11 @@ public void put(Collection<SinkRecord> sinkRecords) {
         attributes.put(ConnectorUtils.KAFKA_OFFSET_ATTRIBUTE, Long.toString(record.kafkaOffset()));
         attributes.put(ConnectorUtils.KAFKA_TIMESTAMP_ATTRIBUTE, record.timestamp().toString());
       }
+      if (includeHeaders && record.headers() != null && !record.headers().isEmpty()) {
+        for (Header header : record.headers()) {
+          attributes.put(header.key(), header.value().toString());
```
kamalaboulhosn: Note that Cloud Pub/Sub limits one to 100 attributes with a key of up to 256 bytes and a value of up to 1024 bytes. Could the headers that come from Kafka exceed these limits?

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java
```diff
@@ -169,40 +171,26 @@ public void start(Map<String, String> props) {
...
           for (Entry<String, String> attribute :
                messageAttributes.entrySet()) {
             if (!attribute.getKey().equals(kafkaMessageKeyAttribute)) {
-              valueSchemaBuilder.field(attribute.getKey(),
-                                       Schema.STRING_SCHEMA);
+              headers.addString(attribute.getKey(), attribute.getValue());
             }
           }
 
-          Schema valueSchema = valueSchemaBuilder.build();
```
kamalaboulhosn: This represents a backwards-incompatible change. I think we should make this a configurable property so that users can choose how the attributes in a message from Cloud Pub/Sub are handled. When false (the default), the struct will be constructed. When true, headers can be used.

cyanezp: I agree with you. It will be included in the next commit.

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
cyanezp: I can't find any information about this kind of restrictions on Kafka's documentation, so I decided to do some quick tests to see how Kafka manages this. Unfortunately, Kafka doesn't complain with any of these three scenarios (more than 100 headers, header with a key over 256 bytes long and header with value over 1024 bytes long) so I'm thinking about putting a new flag to let the user decide how to continue if something like this happens. By default, the message with any of these issues shouldn't be moved to PubSub because it won't be an exact copy of the original message, but maybe the user would like to copy a "truncated" version of the message, so this flag will allow it.

kamalaboulhosn: I don't think dropping the message by default is the right thing to do. That would appear as message loss. Maybe that can be one of the choices, but one should have to make the decision explicitly. Alternatives I can imagine:

1. Truncate headers that exceed the Pub/Sub limits (also lossy, so probably should not be the default).
2. Drop the message entirely. 
3. Put the headers in the STRUCT format of the body of the message and maybe put a special attribute in the message that indicates this is what happened.

The third one seems to be the only way to do this without losing information., so perhaps that should be the default.

cyanezp: I think this feature has becoming a little more complex that I thought. If we look at what is the connector doing right now, there isn't any kafka record headers copying to pubsub attributes, so maybe as a first approach, I can put a disclaimer to indicate that if a kafka record has some headers that are not compatible with pubsub attributes only the message body will be copied (just like is doing now). This way, we can keep discussing how to proceed when we encounter this kind of issues for a future release.

I can't imagine a real usage to include more than a hundred headers on a record, so it may be an extraneous edge case that can be addressed in a later time.

So to recap, if the user enable the "copy attributes" flag we can get 2 scenarios:
 * If all Kafka headers fits with PubSub attributes, then we continue as expected copying everything  without any issue.
 * If some Kafka header does not fit with PubSub attributes or there are more than a hundred attributes, only message's body will be copied into PubSub message without any attributes.

Are you agree with that?

kamalaboulhosn:
Sorry for the delay in looking at this. I have verified the CLA checks manually and will merge. Thanks for the improvements!

PR #209, opened 2019-08-20
DotNet performance test

chrisdunelm:


load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
```diff
@@ -0,0 +1,225 @@
+﻿using CommandLine;
```
dpcollins-google: Please split this up by moving each static class into their own file.  In addition to getting load test results, we'd like to have a relatively short code sample showing users how to configure clients for performance in each language.

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
```diff
@@ -0,0 +1,225 @@
...
+                    var expectedMsgsSent = (now - startTime).TotalSeconds * rate;
+                    if ((rate == 0.0 || msgsSent < expectedMsgsSent) && Interlocked.Add(ref msgsInFlight, 0) < MaxQueued)
+                    {
+                        break;
+                    }
+                    await Task.Delay(TimeSpan.FromMilliseconds(1));
+                }
+                var msgId = msgsSent;
+                var publishTime = DateTime.UtcNow;
+                Array.Copy(BitConverter.GetBytes(msgId), 0, msg, 0, 4);
```
dpcollins-google: Do not use bit bashing.  The api takes attributes, store message id and publish time in those fields using the api https://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.PubSub.V1/api/Google.Cloud.PubSub.V1.PubsubMessage.html

This is what the java client does https://github.com/GoogleCloudPlatform/pubsub/blob/master/load-test-framework/src/main/java/com/google/pubsub/clients/gcloud/CPSPublisherTask.java

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
```diff
@@ -0,0 +1,225 @@
...
+                    var stats = new Stats();
+                    Console.WriteLine($"Starting dotnet server on port:{options.Port}");
+                    var server = new Server
+                    {
+                        Services = { LoadtestWorker.BindService(new LoadtestServer(startTcs, stats.BuildCheckResponse)) },
+                        Ports = { new ServerPort("0.0.0.0", options.Port, ServerCredentials.Insecure) },
+                    };
+                    server.Start();
+                    var startRequest = await startTcs.Task;
+                    Console.WriteLine($"  Start: Test duration: {(int)startRequest.TestDuration.ToTimeSpan().TotalSeconds}s; include ids:{startRequest.IncludeIds}; CPU scaling:{startRequest.CpuScaling}");
```
dpcollins-google: please reformat this to be no wider than 100 columns (java standard)

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
```diff
@@ -0,0 +1,225 @@
...
+                    {
+                        case StartRequest.ClientOptionsOneofCase.PublisherOptions:
+                            await Publish(startRequest, stats);
+                            break;
+                        case StartRequest.ClientOptionsOneofCase.SubscriberOptions:
+                        default: // no option is sent for subscribe
+                            await Subscribe(startRequest, stats);
+                            break;
+                    }
+                    await Task.Delay(TimeSpan.FromHours(6));
```
dpcollins-google: why are you doing this?

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
```diff
@@ -0,0 +1,225 @@
...
+                    _msgIds.Clear();
+                    return ret;
+                }
+            }
+        }
+
+        static async Task<int> Main(string[] args)
+        {
+            var parsed = Parser.Default.ParseArguments<Options>(args);
+            switch (parsed)
```
dpcollins-google: Flatten this out by early exit. 
if (parse failed condition) return 1;
.... parse success code ....

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
dpcollins-google: I should clarify: why are you sleeping for 6 hours here? Isnt there a method on the grpc server to sleep the main thread so the program keeps running while the server is serving?

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
chrisdunelm: done

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
```diff
@@ -0,0 +1,225 @@
...
+        private class Stats
+        {
+            private readonly object _lock = new object();
+            private readonly long _clientId = BitConverter.ToInt64(MD5.Create().ComputeHash(Guid.NewGuid().ToByteArray()));
+            private DateTime? _startTime;
+            private bool _includeIds;
+            private DateTime? _endTime = null;
+
+            private long _errorCount = 0;
+            private long[] _histogram = new long[50];
```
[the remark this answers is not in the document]
chrisdunelm: I feel a fixed-length array is more appropriate. Using a variable-length list would mean checking/extended on every stats record.

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+                    {
+                        stats.End(now);
+                        return;
+                    }
+                    var expectedMsgsSent = (now - startTime).TotalSeconds * rate;
+                    if ((rate == 0.0 || msgsSent < expectedMsgsSent) && Interlocked.Add(ref msgsInFlight, 0) < MaxQueued)
+                    {
+                        break;
+                    }
+                    await Task.Delay(TimeSpan.FromMilliseconds(1));
```
dpcollins-google: Please don't sleep loop here, you have access to the information you need to determine when you have to wait.

For the rate case, you can calculate how long to wait until the next token is available.

For the outstanding case, you can use something like https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim?view=netframework-4.8 with WaitAsync to limit the number outstanding.

load-test-framework/dotnet_src/Google.PubSub.Flic/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+                    await Publish(startRequest, stats);
+                    break;
+                case StartRequest.ClientOptionsOneofCase.SubscriberOptions:
+                default: // no option is sent for subscribe
+                    await Subscribe(startRequest, stats);
+                    break;
+            }
+        }
+
+        private static async Task Publish(StartRequest startRequest, Stats stats)
```
dpcollins-google: Could we move the publisher and subscriber code to separate files, and separate initialization from the action? That way when someone asks "how do I initialize a performant c# publisher", we can point them to the constructor of the publisher, instead of the first section of the publish static method. Something like:

```dotnet
// IWorker.cs
interface IWorker {
  Task Run();
}

// Publisher.cs
internal class Publisher : IWorker {
  public void Init(StartRequest startRequest, Stats stats) {...}
  public async Task Run() {...}
}

// Subscriber.cs
internal class Subscriber : IWorker {
  public void Init(StartRequest startRequest, Stats stats) {...}
  public async Task Run() {...}
}

// Program.cs
IWorker worker;
switch (startRequest.ClientOptionsCase) {
    case StartRequest.ClientOptionsOneofCase.PublisherOptions:
      worker = new Publisher(startRequest, stats);
      break;
    case StartRequest.ClientOptionsOneofCase.SubscriberOptions:
    default:
      worker = new Subscriber(startRequest, stats);
      break;
}
await worker.Run();

```

kamalaboulhosn:
Looks like there hasn't been any follow-up here in some time, so closing.

Issue #206, opened 2019-05-27
PubSub publisher keeps getting DEADLINE_EXCEEDED after a link switchover


piter75:
The other day we had a problem with one of our network links. This particular link is used to route data between one of our data centres and GCP.

The problem with the link started at 10:18PM and  and after tens of seconds the routing has been changed with BGP to use another link and the traffic between our data centre and GCP has been restored.
Nonetheless PubSub connector in our Kafka cluster kept reporting DEADLINE_EXCEEDED errors for another half an hour until 10:48PM when it suddenly started working properly again.

```
May 16 22:48:24 connect-distributed.sh[3366]: [2019-05-16 22:48:24,221] ERROR WorkerSinkTask{id=ohsumi_page_view_raw-0} Commit of offsets threw an unexpected exception for sequence number 40: null (org.apache.kafka.connect.runti
May 16 22:48:24 connect-distributed.sh[3366]: java.lang.RuntimeException: java.util.concurrent.ExecutionException: com.google.api.gax.rpc.DeadlineExceededException: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exc
May 16 22:48:24 connect-distributed.sh[3366]:         at com.google.pubsub.kafka.sink.CloudPubSubSinkTask.flush(CloudPubSubSinkTask.java:296)
May 16 22:48:24 connect-distributed.sh[3366]:         at org.apache.kafka.connect.sink.SinkTask.preCommit(SinkTask.java:125)
May 16 22:48:24 connect-distributed.sh[3366]:         at org.apache.kafka.connect.runtime.WorkerSinkTask.commitOffsets(WorkerSinkTask.java:379)
May 16 22:48:24 connect-distributed.sh[3366]:         at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:209)
May 16 22:48:24 connect-distributed.sh[3366]:         at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:193)
May 16 22:48:24 connect-distributed.sh[3366]:         at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:175)
May 16 22:48:24 connect-distributed.sh[3366]:         at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:219)
[...]

PR #212, opened 2019-11-08
Set keepAlive setting of GRPC channel to 60s to prevent hanging.

samarthsingal:


kamalaboulhosn:
We can do this, but note that I don't think this won't help #206. That uses the Sink connector, which uses the standard Cloud Pub/Sub library. This does not call getChannel at all.

samarthsingal:
Agreed. I  mis-read the code.

PR #223, opened 2020-05-21
fix: Handle null message gracefully.

### LINKED ISSUE #219: NPE for "null" messages on kafka topics
Under certain conditions (using Debezium with SQL Server + Kafka to PubSub Sink), it's possible to generate a "null" message that causes NPE and shutdown of connector. 

Below is a snippet of the exception:

```
RROR WorkerSinkTask{id=CPSSinkConnector-4} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted. Error: null (org.apache.kafka.connect.runtime.WorkerSinkTask)

java.lang.NullPointerException
	at com.google.protobuf.ByteString.copyFromUtf8(ByteString.java:400)
	at com.google.pubsub.kafka.sink.CloudPubSubSinkTask.handleValue(CloudPubSubSinkTask.java:232)
	at com.google.pubsub.kafka.sink.CloudPubSubSinkTask.put(CloudPubSubSinkTask.ja
```
```

### PR DESCRIPTION
**Pull request by kamalaboulhosn**
Fixes #219

### CODE REVIEW

File: kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceTask.java [hunk 1]
```diff
@@ -279,23 +280,21 @@ private void ackMessages() {
...
-          });
+      final ApiFuture<Empty> response = subscriber.ackMessages(requestBuilder.build());
+      response.addListener(new Runnable() {
+        @Override
+        public void run() {
+          try {
+            response.get();
+            log.trace("Successfully acked a set of messages. {}", ackIdsBatch.size());
+          } catch (Exception e) {
+            deliveredAckIds.addAll(ackIdsBatch);
```
samarthsingal: I find `deliveredAckIds` to be a slightly confusing name in this context since they are pending ack ids to be acked to the server. Perhaps we can re-name to `receivedAckIds`?

kamalaboulhosn: That's from a previous PR that was already merged in and so it isn't actually part of this change. Looks like the view might have been set to look at the changes for a single commit?

PR #246, opened 2020-11-24
feat: Add ability to specify endpoint in CPS connector.

kamalaboulhosn:


kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubGRPCSubscriber.java
```diff
@@ -41,9 +41,11 @@
   private GrpcSubscriberStub subscriber;
   private Random rand = new Random(System.currentTimeMillis());
   private CredentialsProvider gcpCredentialsProvider;
+  private String endpoint;
 
-  CloudPubSubGRPCSubscriber(CredentialsProvider gcpCredentialsProvider) {
+  CloudPubSubGRPCSubscriber(CredentialsProvider gcpCredentialsProvider, String endpoint) {
```
dpcollins-google, line 46: nit. @Nullable since this is rare for strings, or consider making "pubsub.googleapis.com" the default flag value and not thinking about null?

dpcollins-google (approved):

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubGRPCSubscriber.java
kamalaboulhosn: Went with a default flag value.

PR #263, opened 2021-03-09
feat: Implement Streaming pull CloudPubSubSubscriber

dpcollins-google:
Also rework interfaces to not be tightly coupled with pull requests.

The implementation here is largely copied from the Pub/Sub Lite BlockingPullSubscriberImpl (https://github.com/googleapis/java-pubsublite/blob/master/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/BlockingPullSubscriberImpl.java) with modifications made to make it work with Cloud Pub/Sub clients instead of Pub/Sub lite wire clients.

### DIFF
--- a/kafka-connector/pom.xml
+++ b/kafka-connector/pom.xml
@@ -7,9 +7,9 @@
   <name>Archetype - kafka-connector</name>
   <url>http://maven.apache.org</url>
   <properties>
-    <maven.compiler.source>1.7</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
-</properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+  </properties>
 
   <dependencies>
     <dependency>
--- a/kafka-connector/pom.xml
+++ b/kafka-connector/pom.xml
@@ -43,7 +43,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsub</artifactId>
-      <version>1.108.3</version>
+      <version>1.112.0</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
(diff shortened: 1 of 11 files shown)


kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
```diff
@@ -193,6 +198,30 @@ public ConfigDef config() {
...
+        .define(
+            CPS_STREAMING_PULL_ENABLED,
+            Type.BOOLEAN,
+            false,
+            Importance.MEDIUM,
+            "Whether to use streaming pull for the connector to connect to Cloud Pub/Sub. If provided, cps.maxBatchSize is ignored.")
+        .define(
+            CPS_STREAMING_PULL_PARALLEL_PULL_COUNT,
+            Type.INT,
+            10,
```
kamalaboulhosn, line 210: Is it best to achieve parallelism through the parallel pull count or by increasing the number of tasks. Having a high parallel pull count can result in worse performance at times, I believe due to contention. We default the client to a single stream and at the very least, I think we should keep that the default. But we should also consider if we even need this knob given the ability to control the number of tasks.

kafka-connector/pom.xml
```diff
@@ -7,9 +7,9 @@
   <name>Archetype - kafka-connector</name>
   <url>http://maven.apache.org</url>
   <properties>
-    <maven.compiler.source>1.7</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
-</properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
```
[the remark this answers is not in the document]
dpcollins-google: Kafka connect dropped java 7 2 years ago https://docs.confluent.io/platform/current/installation/versions-interoperability.html#java so there's no reason we should still have to support it. Its only backwards incompatible if you're still running your cluster with java7

kafka-connector/src/main/java/com/google/pubsub/kafka/source/CloudPubSubSourceConnector.java
dpcollins-google: I worry that it will be high overhead to have a bunch of 5MiB/s tasks? But I can remove this if task fanout is the preferred way to scale.

dpcollins-google: Removed.

kamalaboulhosn: I think we can hit up to 10MB/s per stream. So let's leave this out for now as you have, but we can always add it in later if the overhead of each task is too much and the bottleneck in receiving more messages is the single stream.

kamalaboulhosn (approved):

PR #274, opened 2021-04-29
Add new parameters to source connector 

dpcollins-google:
These control num pullers, max ack extension period and max duration per ack extension.


dpcollins-google:
continuous-integration/travis-ci/pr

kamalaboulhosn (approved):

PR #282, opened 2021-06-17
Add blocking flow controls to the PubSub Sink connector

jordanbull:
Allowing the flow controls to limit the max outstanding bytes and elements. During large traffic spikes, we have observed Connect playing messages into the PubSub producer client as fast as possible resulting in several GB of off heap usage causing OOM errors. Configuring the connector to block on maxOutstandingRequestBytes should prevent this and stop polling until space is available.

We've tested this both with  the default and with a specified low maxOutstandingRequestBytes in our loadtest environment and observed it working as expected.

I've set the default values to Long. MAX_VALUE to effectively preserve the existing behavior for anyone that may not want this change, but I'm open to setting these to a reasonable value if we have some consensus on what that is.

### DIFF
--- a/kafka-connector/README.md
+++ b/kafka-connector/README.md
@@ -124,6 +124,8 @@ Connector supports the following configs:
 | cps.endpoint | String | "pubsub.googleapis.com:443" | The [Cloud Pub/Sub endpoint](https://cloud.google.com/pubsub/docs/reference/service_apis_overview#service_endpoints) to use. |
 | maxBufferSize | Integer | 100 | The maximum number of messages that can be received for the messages on a topic partition before publishing them to Cloud Pub/Sub. |
 | maxBufferBytes | Long | 10000000 | The maximum number of bytes that can be received for the messages on a topic partition before publishing them to Cloud Pub/Sub. |
+| maxOutstandingRequestBytes | Long | Long.MAX_VALUE | The maximum number of outstanding bytes for incomplete and unsent messages before the publisher will block further publishing. |
+| maxOutstandingMessages | Long | Long.MAX_VALUE | The maximum number of incomplete and unsent messages before the publisher will block further publishing. |
 | maxDelayThresholdMs | Integer | 100 | The maximum amount of time to wait to reach maxBufferSize or maxBufferBytes before publishing outstanding messages to Cloud Pub/Sub. |
 | maxRequestTimeoutMs | Integer | 10000 | The timeout for individual publish requests to Cloud Pub/Sub. |
 | maxTotalTimeoutMs | Integer | 60000| The total timeout for a call to publish (including retries) to Cloud Pub/Sub. |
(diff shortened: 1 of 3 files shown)


jordanbull:
@googlebot I signed it!

jordanbull:
@googlebot I signed it!

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -392,6 +401,11 @@ private void createPublisher() {
                     .setDelayThreshold(Duration.ofMillis(maxDelayThresholdMs))
                     .setElementCountThreshold(maxBufferSize)
                     .setRequestByteThreshold(maxBufferBytes)
+                    .setFlowControlSettings(FlowControlSettings.newBuilder()
```
samarthsingal, line 404: Can we only populate these if the maxOutstandingRequestBytes and maxOutstanding{ElementCount,Messages} is less than the default otherwise we'll unnecessarily slow down publishing by acquiring/releasing flow controller locks https://github.com/googleapis/java-pubsub/blob/0bba01661c01ca540625f63a0d7e89d43e2dab34/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java#L247-L257.

samarthsingal (changes requested):
Thanks for contributing this change! We were in the process of doing this ourselves.

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
jordanbull: Sure

jordanbull:
Made the requested changes and did a squash rebase

kafka-connector/README.md
```diff
@@ -124,6 +124,8 @@ Connector supports the following configs:
 | cps.endpoint | String | "pubsub.googleapis.com:443" | The [Cloud Pub/Sub endpoint](https://cloud.google.com/pubsub/docs/reference/service_apis_overview#service_endpoints) to use. |
 | maxBufferSize | Integer | 100 | The maximum number of messages that can be received for the messages on a topic partition before publishing them to Cloud Pub/Sub. |
 | maxBufferBytes | Long | 10000000 | The maximum number of bytes that can be received for the messages on a topic partition before publishing them to Cloud Pub/Sub. |
+| maxOutstandingRequestBytes | Long | Long.MAX_VALUE | The maximum number of outstanding bytes for incomplete and unsent messages before the publisher will block further publishing. |
```
samarthsingal, line 127: nit: s/incomplete/outstanding/g

kafka-connector/README.md
```diff
@@ -124,6 +124,8 @@ Connector supports the following configs:
...
+| maxOutstandingMessages | Long | Long.MAX_VALUE | The maximum number of incomplete and unsent messages before the publisher will block further publishing. |
```
samarthsingal, line 128: nit: s/incomplete/outstanding/g

samarthsingal (approved):
Looks great, thanks!

PR #283, opened 2021-06-17
Sink Connector should publish all outstanding messages on flush

jordanbull:
When flush is called with a non-full batch, the publisher client may wait up to maxDelayThresholdMs before publishing the remaining messages introducing unnecessary tail latency on every flush. This change will force those messages to be published at the start of the flush call.

### DIFF
--- a/kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
+++ b/kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
@@ -334,6 +334,8 @@ public class CloudPubSubSinkTask extends SinkTask {
 
   @Override
   public void flush(Map<TopicPartition, OffsetAndMetadata> partitionOffsets) {
+    // Publish the incomplete batch now instead of waiting for the maxDelayThresholdMs
+    publisher.publishAllOutstanding();
     log.debug("Flushing...");
     // Process results of all the outstanding futures specified by each TopicPartition.
     for (Map.Entry<TopicPartition, OffsetAndMetadata> partitionOffset :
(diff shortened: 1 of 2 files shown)


jordanbull:
@googlebot I signed it!

samarthsingal:
Can you elaborate why this is needed? As implemented the flush() call just processes the results of the publishing and the publisher does it own batching. flush() does not cause any publishing. With this, batching will be difficult to predict since it will at both flush time and when maxDelayThresholdMs elapses. If tail latency is a concern, have you tried just using a lower setting for maxDelayThresholdMs?

jordanbull:
Given Connect executes both the flush/commit and polls/puts in a synchronous loop, there will be no new messages arriving while waiting for flush to complete. Flush currently is awaiting completion of every message put in the producer anyways. Since there aren't any new messages coming, I figured we might as well send the batch immediately rather than wait for the time to elapse given the number of messages in that batch isn't going to change.

samarthsingal:
My understanding was that poll/flush cycle was async and the sink would keep polling and delivering more messages regardless of whether the previous put() call had committed. That is my reading from https://github.com/a0x8o/kafka/blob/93185ee0de2349c7b8413f48611f3ce1e10dc951/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L215-L232- in each iteration() call it first tries to commit and then poll for new messages. 

Is it documented that this is synchronous?

jordanbull:
> My understanding was that poll/flush cycle was async and the sink would keep polling and delivering more messages regardless of whether the previous put() call had committed. That is my reading from https://github.com/a0x8o/kafka/blob/93185ee0de2349c7b8413f48611f3ce1e10dc951/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L215-L232- in each iteration() call it first tries to commit and then poll for new messages.
> 
> Is it documented that this is synchronous?

Unfortunately, I'm unable to find documentation that explicitly states that (maybe worth contributing to the java docs in connect), but as I read it, both the WorkerSinkTask code and the flush method of the PubSubSinkTask already depend on this fact for correctness.

IE PubSubSinkTask.flush would have a race in awaiting the future results and clearing them at https://github.com/GoogleCloudPlatform/pubsub/blob/master/kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java#L369

Similarly in WorkerSinkTask polling adds offsets to the un-synchronized currentOffsets map and then the commit call clears that same map https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L571

PR #294, opened 2021-07-29
fix: use ubuntu-2004-focal-v20210720 as GCE VM source image

### LINKED ISSUE #293: GCEComputeResourceController creates VM incompatible default Python version
The `GCEComputeResourceController` class creates a GCE VM whose default Python version is incompatible with the client source code.

This line https://github.com/GoogleCloudPlatform/pubsub/blob/6fbb5fbf561a7021a1564601948e371f3db4e4fe/load-test-framework/src/main/java/com/google/pubsub/flic/controllers/resource_controllers/GCEComputeResourceController.java#L37-L38 creates a GCE VM with Ubuntu 16.04 Xenial. The default Python 3 version is 3.5.2.

This client code https://github.com/GoogleCloudPlatform/pubsub/blob/6fbb5fbf561a7021a1564601948e371f3db4e4fe/load-test-framework/python_src/clients/cps_publisher_task.py#L35-L39 uses variable annotations which aren't supported until Python 3.6 \[1\].

The `SOURCE_FAMILY` should probably be changed to use an Ubuntu version whose default Python is >= 3.6 like Focal 20.04.

1: https://www.python.org/dev/peps/pep-0526/

davidxia:
fixes #293

### DIFF
--- a/load-test-framework/src/main/java/com/google/pubsub/flic/controllers/resource_controllers/GCEComputeResourceController.java
+++ b/load-test-framework/src/main/java/com/google/pubsub/flic/controllers/resource_controllers/GCEComputeResourceController.java
@@ -35,7 +35,7 @@ import org.slf4j.LoggerFactory;
 public class GCEComputeResourceController extends ComputeResourceController {
   private static final String MACHINE_TYPE = "n1-standard-"; // standard machine prefix
   private static final String SOURCE_FAMILY =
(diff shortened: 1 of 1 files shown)


davidxia:
cc @samarthsingal

davidxia:
cc @dpcollins-google if you can review. 🙏

dpcollins-google (approved):

PR #298, opened 2021-10-15
Update copy_tool.py

hicod3r:
Modified the download link/URL pattern for KAFKA_LINK to point to the latest version. The previous pattern/version was broken.

### DIFF
--- a/kafka-connector/copy_tool.py
+++ b/kafka-connector/copy_tool.py
@@ -17,9 +17,9 @@ import tarfile
 import tempfile
 import subprocess
 
-KAFKA_RELEASE = "2.6.0"
+KAFKA_RELEASE = "3.0.0"
 KAFKA_FOLDER = f"kafka_2.13-{KAFKA_RELEASE}"
-KAFKA_LINK = f"https://downloads.apache.org/kafka/{KAFKA_RELEASE}/{KAFKA_FOLDER}.tgz"
+KAFKA_LINK = f"https://dlcdn.apache.org/kafka/{KAFKA_RELEASE}/{KAFKA_FOLDER}.tgz"
 
 CONNECTOR_RELEASE = "v0.10-alpha"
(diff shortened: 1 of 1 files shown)


kafka-connector/copy_tool.py
```diff
@@ -17,9 +17,9 @@
 import tempfile
 import subprocess
 
-KAFKA_RELEASE = "2.6.0"
+KAFKA_RELEASE = "3.0.0"
 KAFKA_FOLDER = f"kafka_2.13-{KAFKA_RELEASE}"
-KAFKA_LINK = f"https://downloads.apache.org/kafka/{KAFKA_RELEASE}/{KAFKA_FOLDER}.tgz"
+KAFKA_LINK = f"https://dlcdn.apache.org/kafka/{KAFKA_RELEASE}/{KAFKA_FOLDER}.tgz"
```
samarthsingal, line 22: We can change the link to https://archive.apache.org/dist/kafka/2.6.0/kafka_2.13-2.6.0.tgz instead.

kafka-connector/copy_tool.py
```diff
@@ -17,9 +17,9 @@
 import tempfile
 import subprocess
 
-KAFKA_RELEASE = "2.6.0"
+KAFKA_RELEASE = "3.0.0"
```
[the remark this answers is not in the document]
samarthsingal: This is changing a major version that has many changes, including API changes and dropped complatibility. for Java 8 - https://downloads.apache.org/kafka/3.0.0/RELEASE_NOTES.html. We havent fully tested with Kafka 3.0 and will defer this upgrade until then.

Lets revert to 2.6.0.

samarthsingal (changes requested):

kafka-connector/copy_tool.py
ajprabhakar-google: Sounds good. We can change it to the archive link instead

PR #304, opened 2021-12-16
Remove log4j

kamalaboulhosn:


amfisher-404 (approved):
Per our discussion, removing logging is a better present alternative than keeping the vulnerability.

PR #306, opened 2021-12-27
PubSubSinkTask.flush should flush tracking for all input partitions and only those partitions

jordanbull:
The existing behavior will only clear tracking for the first partition that throws an exception in flush. In cases where the partitions complete successfully, it will also clear tracking for partitions that it was not asked to flush. Both of these are not the expected behavior when implementing the interface. This pull requests fixes these two behaviors and adds tests that catch them in the old code and pass with the changes.

### DIFF
--- a/kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
+++ b/kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
@@ -347,6 +347,7 @@ public class CloudPubSubSinkTask extends SinkTask {
   @Override
   public void flush(Map<TopicPartition, OffsetAndMetadata> partitionOffsets) {
     log.debug("Flushing...");
+    RuntimeException maybeException = null;
     // Process results of all the outstanding futures specified by each TopicPartition.
     for (Map.Entry<TopicPartition, OffsetAndMetadata> partitionOffset :
         partitionOffsets.entrySet()) {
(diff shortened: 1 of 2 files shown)


kafka-connector/src/test/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTaskTest.java
```diff
@@ -362,18 +364,66 @@ public void testFlushWithNoPublishInPut() throws Exception {
...
   public void testFlushExceptionCase() throws Exception {
     task.start(props);
     Map<TopicPartition, OffsetAndMetadata> partitionOffsets = new HashMap<>();
     partitionOffsets.put(new TopicPartition(KAFKA_TOPIC, 0), null);
-    List<SinkRecord> records = getSampleRecords();
+    List<SinkRecord> records = getSampleRecords(0);
     ApiFuture<String> badFuture = getFailedPublishFuture();
     when(publisher.publish(any(PubsubMessage.class))).thenReturn(badFuture);
     task.put(records);
+    assertThrows(RuntimeException.class, () -> task.flush(partitionOffsets));
```
jordanbull, line 376: Use assertThrows instead of expected so the following verifies are actually reached.

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -356,20 +357,29 @@ public void flush(Map<TopicPartition, OffsetAndMetadata> partitionOffsets) {
...
+      int partition = partitionOffset.getKey().partition();
       OutstandingFuturesForPartition outstandingFutures =
-          outstandingFuturesForTopic.get(partitionOffset.getKey().partition());
+          outstandingFuturesForTopic.get(partition);
       if (outstandingFutures == null) {
         continue;
       }
       try {
-        ApiFutures.allAsList(outstandingFutures.futures).get();
+        // Only wait for partition to complete if the clush hasn't already failed.
```
samarthsingal, line 367: typo

samarthsingal: "clush"

kafka-connector/src/main/java/com/google/pubsub/kafka/sink/CloudPubSubSinkTask.java
```diff
@@ -356,20 +357,29 @@ public void flush(Map<TopicPartition, OffsetAndMetadata> partitionOffsets) {
...
+        if (maybeException == null) {
```
samarthsingal, line 368: I'm a little confused by this. If we have multiple partitions, one of which fails, we dont get the future for the rest but do clear the tracking? The publish could still be outstanding for the other partitions at the point and so we just wont care about it?

Is the expectation that once `flush` fails for any partition, all partitions will be `put` from the last committed offset?

jordanbull: Yea, either the preCommit/flush succeeds for all the passed in partitions or fails for all of them and the task seeks back to the last committed offsets. https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L410

At that point, we know anything we were tracking for those partitions will get replayed.

samarthsingal: Thanks for confirming.

samarthsingal (approved):

PR #316, opened 2022-06-14
Update release version to 3.2.0

jtamyrc:
URL for release 2.6.0 does not exist anymore.


dpcollins-google:
Hi Jason, thanks for the PR. I've fixed this in a different way with #317 which should be less invasive. Thanks for identifying this issue!

PR #327, opened 2022-10-05
fix(load-test): update startup scripts with new gsutil path

### LINKED ISSUE #323: loadtest: unable to connect to client
I'm trying to run the loadtest framework and there seems to be an issue with connecting to the GCE client.

Here's the command I'm running

`python3 run.py --project=alxh-test --language=GO --test_parameters=latency`

After a while, the program times out with

```
Sep 22, 2022 12:11:04 AM com.google.common.util.concurrent.AbstractFuture executeListener
SEVERE: RuntimeException while executing runnable CallbackListener{com.google.pubsub.flic.controllers.resource_controllers.GCEComputeResourceController$1@69aa771b} with executor MoreExecutors.directExecutor()
java.lang.RuntimeException: Unable to connect to client <redacted>:5000 in 600 seconds.
```

The firewall rule seems to be being applied properly and there are lines that indicate the `InstanceGroupManager` is ready, though individual instances aren't.

alvarowolfx:
Partially resolves #323, as the startup scripts on the VMs were failing due to the wrong `gsutil` path.


dpcollins-google (approved):

alvarowolfx:
@dpcollins-google I don't have permissions to merge this. Can we merge before it gets stale ?

PR #343, opened 2023-05-30
feat: create pubsub flink source

hannahrogers-google:


flink-connector/src/main/java/com/google/pubsub/flink/PubSubSource.java
```diff
@@ -0,0 +1,158 @@
...
+import org.apache.flink.api.connector.source.SplitEnumerator;
+import org.apache.flink.api.connector.source.SplitEnumeratorContext;
+import org.apache.flink.api.java.typeutils.ResultTypeQueryable;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.core.io.SimpleVersionedSerializer;
+import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.util.UserCodeClassLoader;
+
+@AutoValue
+public abstract class PubSubSource<OutputT>
```
dpcollins-google, line 51: Please split the source from its setting object, like PubSubLiteSource

dpcollins-google (approved):

flink-connector/src/main/java/com/google/pubsub/flink/PubSubSource.java
hannahrogers-google: My preference would be to keep this as is because the plan is to migrate this connector to  https://github.com/apache/flink-connector-gcp-pubsub to replace the apache owned connector.

Having the builder apart of Source object is more consistent with the existing connector (and other existing connectors in the apache repo).

PR #356, opened 2024-02-26
Add flow control setting to source builder

matt-kwong:
Default flow control values are copied from the CPS client library. `FlowControlSettings` cannot be added as a builder option due to it not being serializable.

Max outstanding values are per-checkpoint. I plan to document how flow control works in a separate PR.

### DIFF
--- a/flink-connector/flink-connector-gcp-pubsub/src/test/java/com/google/pubsub/flink/PubSubSourceTest.java
+++ b/flink-connector/flink-connector-gcp-pubsub/src/test/java/com/google/pubsub/flink/PubSubSourceTest.java
@@ -89,13 +89,6 @@ public class PubSubSourceTest {
   }
 
   @Test
-  public void build_invalidLimitExceededBehavior() throws Exception {
-    assertThrows(
-        NullPointerException.class,
-        () -> PubSubSource.<String>builder().setLimitExceededBehavior(null));
-  }
-
-  @Test
   public void build_invalidCreds() throws Exception {
     assertThrows(
         NullPointerException.class, () -> PubSubSource.<String>builder().setCredentials(null));
--- a/flink-connector/flink-connector-gcp-pubsub/src/test/java/com/google/pubsub/flink/PubSubSourceTest.java
+++ b/flink-connector/flink-connector-gcp-pubsub/src/test/java/com/google/pubsub/flink/PubSubSourceTest.java
@@ -59,10 +59,14 @@ public class PubSubSourceTest {
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


flink-connector/flink-connector-gcp-pubsub/src/test/java/com/google/pubsub/flink/PubSubSourceTest.java
```diff
@@ -58,6 +58,27 @@ public void build_invalidSchema() throws Exception {
     assertThrows(IllegalStateException.class, builder::build);
   }
 
+  @Test
+  public void build_invalidMaxOutstandingMessagesCount() throws Exception {
```
clmccart, line 62: do we want to throw an exception on negative values? (here and below)

clmccart (changes requested):

flink-connector/flink-connector-gcp-pubsub/src/test/java/com/google/pubsub/flink/PubSubSourceTest.java
matt-kwong: Good idea! I added throwing an exception on negative or zero values. This mimics the behavior of the underlying library that uses these settings. Surfacing the exception earlier, when building `PubSubSource`, will make debugging failing Flink jobs much easier.

matt-kwong:
I removed the `limitExceededBehavior` option, since subscribers don't actually use that field.

flink-connector/flink-connector-gcp-pubsub/src/test/java/com/google/pubsub/flink/PubSubSourceTest.java
```diff
@@ -58,6 +58,36 @@ public void build_invalidSchema() throws Exception {
...
+    assertThrows(
+        NullPointerException.class,
+        () -> PubSubSource.<String>builder().setMaxOutstandingMessagesCount(null));
+    PubSubSource.Builder<String> builder =
```
clmccart, line 66: this feels like two different tests to me. one for NullPointerException and one for IllegalArgumentException. (same below)

clmccart (changes requested):

flink-connector/flink-connector-gcp-pubsub/src/test/java/com/google/pubsub/flink/PubSubSourceTest.java
matt-kwong: Created separate test cases

clmccart (approved):

PR #361, opened 2024-03-21
Add builder options to source/sink

matt-kwong:
These are commonly used builder options in the underlying Pub/Sub client library.

### DIFF
--- a/flink-connector/flink-connector-gcp-pubsub/src/main/java/com/google/pubsub/flink/PubSubSink.java
+++ b/flink-connector/flink-connector-gcp-pubsub/src/main/java/com/google/pubsub/flink/PubSubSink.java
@@ -38,6 +38,10 @@ public abstract class PubSubSink<T> implements Sink<T> {
 
   public abstract Optional<Credentials> credentials();
 
+  public abstract Optional<Boolean> enableMessageOrdering();
+
+  public abstract Optional<String> endpoint();
+
   public static <T> Builder<T> builder() {
     return new AutoValue_PubSubSink.Builder<T>();
   }
@@ -46,6 +50,12 @@ public abstract class PubSubSink<T> implements Sink<T> {
     Publisher.Builder builder = Publisher.newBuilder(topicName.toString());
     if (credentials().isPresent()) {
       builder.setCredentialsProvider(FixedCredentialsProvider.create(credentials().get()));
+    }
+    if (enableMessageOrdering().isPresent()) {
+      builder.setEnableMessageOrdering(enableMessageOrdering().get());
+    }
+    if (endpoint().isPresent()) {
+      builder.setEndpoint(endpoint().get());
     }
     return builder.build();
   }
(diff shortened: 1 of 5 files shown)


flink-connector/flink-connector-gcp-pubsub/src/main/java/com/google/pubsub/flink/PubSubSink.java
```diff
@@ -75,6 +85,10 @@ public abstract Builder<T> setSerializationSchema(
 
     public abstract Builder<T> setCredentials(Credentials credentials);
 
+    public abstract Builder<T> setEnableMessageOrdering(Boolean enableMessageOrdering);
```
clmccart, line 88: if a caller sets these values, they'll only be applied when a new publisher is created, correct? if so, we should document that caveat because the the name of the setter kinda implies an update

clmccart: same with the subscriber

flink-connector/flink-connector-gcp-pubsub/src/main/java/com/google/pubsub/flink/PubSubSink.java
```diff
@@ -38,6 +38,10 @@ public abstract class PubSubSink<T> implements Sink<T> {
 
   public abstract Optional<Credentials> credentials();
 
+  public abstract Optional<Boolean> enableMessageOrdering();
```
clmccart, line 41: can we document these options (and the other ones in this file)?

clmccart (changes requested):

flink-connector/flink-connector-gcp-pubsub/src/main/java/com/google/pubsub/flink/PubSubSink.java
matt-kwong: This method is in PubSubSink.Builder, so it can only be called before any publishers are created and cannot be used to update Pub/Sub sink behavior after it has started (at least from my understanding of Flink).

This is the same naming convention used by the underlying CPS client library, which uses a builder pattern to configure immutable properties for publisher/subscriber clients. Do you have any naming suggestions to clarify the intent?

I agree with documenting behavior/expectations.

clmccart: yeah that's my understanding as well. i think leaving the naming as is is fine and just adding a comment describing the behavior

flink-connector/flink-connector-gcp-pubsub/src/main/java/com/google/pubsub/flink/PubSubSink.java
matt-kwong: Done.

flink-connector/flink-connector-gcp-pubsub/src/main/java/com/google/pubsub/flink/PubSubSink.java
matt-kwong: Done.

clmccart (approved):

PR #379, opened 2025-04-11
simply make auto-value works

bindiego:
It doesn't compile without this. Do I miss anything?

`mvn clean package -DskipTests`


matt-kwong (changes requested):
Sorry for the delayed review! Thanks for making this fix!

Could you remove `flink-connector/flink-connector-gcp-pubsub/dependency-reduced-pom.xml` from the PR and add `dependency-reduced-pom.xml` to the `.gitignore` file?

matt-kwong (approved):

PR #411, opened 2026-04-06
Replace ordering key prober's --should_cleanup with --should_not_cleanup so that we can use the flag

TimeString:
The current implementation will not be able to disable the flag (i.e., specify that I don't want to cleanup). This can be resolved by adding `arity=1` in the @parameter() (see https://jcommander.org/ section 2.1), but I try to keep the style consistent with `--no_publish` flag, I decide to just introduce the negation in the flag name.

### DIFF
--- a/ordering-keys-prober/src/main/java/com/google/cloud/pubsub/prober/ProberStarter.java
+++ b/ordering-keys-prober/src/main/java/com/google/cloud/pubsub/prober/ProberStarter.java
@@ -40,7 +40,7 @@ public class ProberStarter {
     @Parameter(names = "--endpoint", description = "Cloud Pub/Sub endpoint to run against.")
     private String endpoint = "pubsub.googleapis.com:443";
 
-    @Parameter(names = "--should_not_cleanup", description = "Whether the prober should start by not cleaning up the topic and subscription.")
+    @Parameter(names = "--no_cleanup", description = "Skip the cleanup of the topic and subscription on startup.")
     private boolean shouldNotCleanup = false;
 
     @Parameter(names = "--no_publish", description ="Whether the prober should skip publishing.")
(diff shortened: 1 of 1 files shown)


ordering-keys-prober/src/main/java/com/google/cloud/pubsub/prober/ProberStarter.java
```diff
@@ -40,8 +40,8 @@ public static class Args {
     @Parameter(names = "--endpoint", description = "Cloud Pub/Sub endpoint to run against.")
     private String endpoint = "pubsub.googleapis.com:443";
 
-    @Parameter(names = "--should_cleanup", description ="Whether the prober should start by cleaning up the topic and subscription.")
-    private boolean shouldCleanup = true;
+    @Parameter(names = "--should_not_cleanup", description = "Whether the prober should start by not cleaning up the topic and subscription.")
```
kamalaboulhosn: Can we just call it `no_cleanup`?

kamalaboulhosn: Adding just a "not" in the description is confusing. Let's change it to "Skip the cleanup of the topic and subscription on startup."

kamalaboulhosn (changes requested):

ordering-keys-prober/src/main/java/com/google/cloud/pubsub/prober/ProberStarter.java
TimeString: Done.

TimeString: Revised the flag description per suggestion.

kamalaboulhosn (approved):

